PR #1, opened 2026-01-21
Add initial boilerplate

gndelia:
Adding the initial boilerplate. This creates an app with all the routes. It uses the usual setup, with the only exception is that we are using pnpm as a package manager, and Vite instead of next.

Things missing in this boilerplate:

- knip, as we may add code that's unused until we have something more defined
- workflow actions, as the current ones do not support pnpm


web/src/providers/web3Provider.tsx
```diff
@@ -0,0 +1,21 @@
+import { getDefaultConfig, RainbowKitProvider } from "@rainbow-me/rainbowkit";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { hemi, hemiSepolia } from "viem/chains";
+import { WagmiProvider } from "wagmi";
+
+const config = getDefaultConfig({
+  appName: "Vetro",
+  chains: [hemi, hemiSepolia],
+  // TODO add project id for wallet connect
+  projectId: "YOUR_PROJECT_ID_HERE",
```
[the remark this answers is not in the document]
gndelia: this is just for compilation, we'll get a key eventually

.husky/commit-msg
```diff
@@ -0,0 +1 @@
+pnpm commitlint --edit $1
```
gabmontes: I think you do not need to call `pnpm` in the Husky hooks.

gndelia: Simplified in [5e5fe48](https://github.com/hemilabs/vetro-monorepo/pull/1/commits/5e5fe486536701824327b3b6b9dc51e7a81d52eb)

ArturDolzan (approved):

gabmontes (approved):

PR #5, opened 2026-01-23
Add multi-language support

### LINKED ISSUE #3: Add multilanguage support

With at least English and Spanish, leaving the room for further languages

gndelia:
### Description

This PR adds multilingual support. There's no particular lib for i18n for Vite, so I had to work with React libs. Of those, one of the most popular is `react-i18next` (no, it's not related to the `next` framework). It worked well out of the box, but I had to do a little custom work to enable type-safety and IntelliSense.

Something I found out while testing this is that `vite` by default does not compile with TypeScript (it removes the types [type stripping] when compiling on `vite build` without checking anything - [source](https://vite.dev/guide/features#transpile-only)).

I also added some tests to check the translation files match (taken from [ui-monorepo](https://github.com/hemilabs/ui-monorepo/blob/main/portal/test/locale.test.ts)), and I also added a test that checks that the translation files are sorted (which I will add to the ui-monorepo later)

### Screenshots

https://github.com/user-attachments/assets/39a01be8-eec1-4cf2-badf-5f67c45267e3

I added a network switch for the context of testing, but it won't really live, so do not worry about it. The same goes for the translation texts, they are just for demonstration and not really final texts

### Related issue(s)

Closes #3 

### Checklist

- [x] Manual testing passed.


web/package.json
```diff
@@ -2,29 +2,35 @@
   "name": "@vetro/web",
   "version": "1.0.0",
   "scripts": {
-    "build": "vite build",
+    "build": "tsc && vite build",
     "dev": "vite",
-    "preview": "vite preview"
+    "prepreview": "npm run build",
```
gabmontes: pnpm?

web/src/hooks/useTranslation.ts
```diff
@@ -0,0 +1,35 @@
...
+type ExtractKeys<T, Prefix extends string = ""> = {
+  [K in keyof T]: T[K] extends Record<string, unknown>
+    ? ExtractKeys<T[K], `${Prefix}${K & string}.`>
+    : `${Prefix}${K & string}`;
+}[keyof T];
+
+type TranslationKeys = ExtractKeys<typeof resources.en.translation>;
+
+/**
+ * Custom wrapper around react-i18next's useTranslation hook that provides strict type safety.
```
gabmontes: Is this possible by patching the lib instead?

web/src/i18n/config.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { resources } from "./resources";
+
+export const initializeI18n = () =>
+  i18n.use(initReactI18next).init({
+    debug: import.meta.env.DEV,
+    fallbackLng: "en",
+    lng: "en", // Default language
+    resources,
+    supportedLngs: ["en", "es"],
```
gabmontes, line 14: Ask Akash & co. what languages we should support. I think English and Chinese. Spanish is for free.

gndelia: Yes, I just added English and Spanish as placeholders. We'll later start adding in the appropriate languages

web/package.json
gndelia: Updated in [85fd72c](https://github.com/vetro-protocol/vetro-monorepo/pull/5/commits/85fd72c78b3af8c8922aff75eb70adca5b35c80b). Though using npm does work too 😅

web/src/hooks/useTranslation.ts
gndelia: It can't be solved with patching because the type definition includes the translation files, which come from the web project. I'll see if I can improve this a bit, though, because the support for these libs is a bit more limited than the ones in next

gndelia: I found a better way without a custom wrapper at [8235ab9](https://github.com/vetro-protocol/vetro-monorepo/pull/5/commits/8235ab905c8a3eb3395777edcbce10aa8bff6adc)

ArturDolzan (approved):

gabmontes (approved):

Issue #6, opened 2026-01-26
Implement the Swap page


gndelia:

gndelia:
This was mostly completed. Improvements will be done in subsequent issues

PR #14, opened 2026-01-29
Refactor address validation+get treasury

gndelia:
### Description

This PR adds the method to get the treasury address (which will be needed for getting price deviation in swap operations + the list of whitelisted tokens), as well as creating a `isValidAddress` utility to simplify the checks for addresses (which were repeated in many places)

### Screenshots

N/A

### Related issue(s)

Related to #6

### Checklist

- [x] Manual testing passed.


packages/gateway/src/utils/isAddressValid.ts
```diff
@@ -0,0 +1,4 @@
+import { type Address, isAddress, isAddressEqual, zeroAddress } from "viem";
+
```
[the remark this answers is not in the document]
gndelia: not needed - it's an internal function and pretty clear to read

arielmergen (dismissed):

packages/gateway/src/utils/isAddressValid.ts
```diff
@@ -0,0 +1,4 @@
...
+export const isAddressValid = (address: Address | undefined) =>
+  address && isAddress(address) && !isAddressEqual(address, zeroAddress);
```
ArturDolzan: It looks like `isAddressValid` returns undefined when address is undefined. Is this the expected behavior? Otherwise, you could explicitly define it to return a boolean... something like:
```suggestion
export const isAddressValid = (address: Address | undefined): boolean =>
    !!address && isAddress(address) && !isAddressEqual(address, zeroAddress);
```

ArturDolzan (commented):
I left a minor comment otherwise LGTM.

packages/gateway/src/utils/isAddressValid.ts
gndelia: You're correct - let me fix this

gndelia: Fixed in [c3a84d4](https://github.com/vetro-protocol/vetro-monorepo/pull/14/commits/c3a84d480964b7f90ab9802d710e5e04d5c92f1b)

ArturDolzan (approved):

gabmontes (approved):

PR #35, opened 2026-02-04
Add EarnStats section with staking balance hooks

ArturDolzan:
### Description

This PR implements:

- Create EarnStats component with three StatCards
- Add useStakedBalance and useEarnedAmount hooks
- Add fetchStakedBalance fetcher for ERC4626 vault
- Move StatCard to components folder
- Update translations for earn page stats
- Add viem-erc4626 and react-loading-skeleton deps

### Screenshots

https://github.com/user-attachments/assets/d727d350-1a7e-48aa-8838-ee46b41df0f2

https://github.com/user-attachments/assets/3d6799c7-ac93-443b-834a-db5d0f9f2b3f

### Related issue(s)

Related to #23

### Checklist

- [x] Manual testing passed.


web/src/hooks/useStakedBalance.ts
```diff
@@ -0,0 +1,41 @@
...
+  const chain = useMainnet();
+  const client = useEthereumClient();
+  const stakingVaultAddress = getStakingVaultAddress(chain.id);
+
+  return useQuery({
+    enabled: !!client && !!account,
+    queryFn: () =>
+      fetchStakedBalance({
+        account: account!,
+        client: client!,
```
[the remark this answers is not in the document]
ArturDolzan: As you mentioned, line 28 already ensures these values are defined.

web/src/i18n/locales/es/translation.json
```diff
@@ -19,7 +19,12 @@
       "title": "Hola Bridge"
     },
     "earn": {
-      "title": "Hola Earn"
+      "stats": {
+        "earnedAmount": "Tu monto ganado",
+        "rewards": "Tus recompensas",
+        "stakedBalance": "Tu balance stakeado"
```
gabmontes:
```suggestion
        "earnedAmount": "Su monto ganado",
        "rewards": "Sus recompensas",
        "stakedBalance": "Su balance stakeado"
```

web/src/i18n/locales/es/translation.json
```diff
@@ -19,7 +19,12 @@
...
+      },
+      "title": "Stake VUSD para ganar rendimiento variable."
```
gabmontes:
```suggestion
      "title": "Haga stake de VUSD para ganar un rendimiento variable."
```

web/src/pages/earn/components/earnStats/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+        label={t("pages.earn.stats.stakedBalance")}
+        value={formatStakedBalance()}
+      />
+      <StatCard
+        icon={<TrendingUpIcon />}
+        isLoading={isLoadingEarnedAmount}
+        label={t("pages.earn.stats.earnedAmount")}
+        value={formatEarnedAmount()}
+      />
+      {/* TODO - Rewards should be implemented as soon we have the API available */}
```
gabmontes, line 56: We have the API but there is no Merlk opportunities or campaigns. Should the API just return i.e. `[]` in this case?

gabmontes (changes requested):

gndelia:
I can't tag Nahuel here 😆  @gabmontes  can we grant him access to this repo?

What I wanted to mention: I am unsure if he has defined how the Loading state should look like (I see you implemented skeletons just like the portal)

web/src/hooks/useEarnedAmount.ts
```diff
@@ -0,0 +1,36 @@
...
+        account: account!,
+        address: stakingVaultAddress,
+      }),
+    queryKey: earnedAmountQueryKey({
+      account: account!,
+      chainId: chain.id,
+      stakingVaultAddress,
+    }),
+  });
+}
```
gndelia: It seems that his hook is just reimplementing the [useTokenBalance](https://github.com/hemilabs/react-hooks/blob/master/src/useTokenBalance.ts)?

After all, an ERC4626 is an extension of ERC20. Would using it work?

web/src/pages/earn/components/earnStats/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const formatStakedBalance = function () {
+    if (stakedBalance !== undefined && vusd) {
+      return `${formatUnits(stakedBalance, vusd.decimals)} ${vusd.symbol}`;
+    }
+    return "";
+  };
+
+  const formatEarnedAmount = function () {
+    if (earnedAmount !== undefined && vusd) {
+      return `${formatUnits(earnedAmount, vusd.decimals)} s${vusd.symbol}`;
```
gndelia: similar as we did on the portal, with `HEMI` vs `veHEMI`,  I don't think we should prepend the `v` to `vusd.symbol` here.

I think we should read the `symbol` from the Earn contract, which is a regular ERC4626 vault. I'm open to implement this in a future issue though, but if that's the case, let's create it now

web/src/pages/earn/components/earnStats/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+export function EarnStats() {
+  const { t } = useTranslation();
+  const { address } = useAccount();
+  const { data: vusd } = useVusd();
+
+  const { data: stakedBalance, isLoading: isLoadingStakedBalance } =
+    useStakedBalance({ account: address });
+
+  const { data: earnedAmount, isLoading: isLoadingEarnedAmount } =
+    useEarnedAmount({ account: address });
```
gndelia: I'm thinking that the `useAccount` could be moved inside of these 2 hooks - I don't think there's a scenario were we used these hooks and a different account needs to be passed. Thoughts?

web/src/fetchers/fetchStakedBalance.ts
```diff
@@ -0,0 +1,24 @@
...
+}: {
+  account: Address;
+  client: Client;
+  stakingVaultAddress: Address;
+}) {
+  // Get user's shares (sVUSD balance)
+  const shares = await balanceOf(client, {
+    account,
+    address: stakingVaultAddress,
+  });
```
gndelia: I think here we could use [ensureQueryData](https://tkdodo.eu/blog/react-query-meets-react-router#a-router-that-fetches-data) to read this info from the cache (check the shared link to understand what's doing).

I'm already using this option [here, for example](https://github.com/vetro-protocol/vetro-monorepo/blob/091704edad3e5da6e8e2858d1a97a3eeba3dba9b/web/src/hooks/useDeposit.ts#L61).

However, for doing so, I think we need to extend [useTokenBalance](https://github.com/hemilabs/react-hooks/blob/master/src/useTokenBalance.ts) to export the balanceOfQueryOptions, similar to [previewDeposit here](https://github.com/vetro-protocol/vetro-monorepo/blob/091704edad3e5da6e8e2858d1a97a3eeba3dba9b/web/src/hooks/usePreviewDeposit.ts#L21)

Would you like to do it? (Btw, not blocking this PR, we can update this here later)

edit: Note that we need to see case-by-case when it makes sense to read from cache and when re-query

web/src/pages/earn/components/earnStats/index.tsx
ArturDolzan: @gabmontes yes, please.

gndelia:
@ArturDolzan  so Nahuel says the Skeleton is ok, but we need to update the gray color to match the gray from our theme

Let's do that, but don't block this PR just for this

web/src/i18n/locales/es/translation.json
ArturDolzan: Applied here 4b16d6c3b8a17cb574a8e989d69c425c651c3a5f. Thanks!

web/src/pages/earn/components/earnStats/index.tsx
ArturDolzan: Can I keep the placeholder to not block this PR and then I'll iterate on upcoming PR's?

web/src/hooks/useEarnedAmount.ts
ArturDolzan: Good point. Simplified here 4b16d6c3b8a17cb574a8e989d69c425c651c3a5f. Thanks for pointing that out.

web/src/pages/earn/components/earnStats/index.tsx
ArturDolzan: Applied here 4b16d6c3b8a17cb574a8e989d69c425c651c3a5f.

web/src/pages/earn/components/earnStats/index.tsx
ArturDolzan: Applied here 4b16d6c3b8a17cb574a8e989d69c425c651c3a5f.

web/src/fetchers/fetchStakedBalance.ts
ArturDolzan: Cool! Applied here 4b16d6c3b8a17cb574a8e989d69c425c651c3a5f.

gndelia (approved):

gabmontes (approved):

PR #85, opened 2026-02-16
api: configure Cloudflare workers

joshuasing:
### Description

Changes required to configure Cloudflare workers for `api/`.

- Add `api/tsconfig.json`
- Switch from `@cloudflare/worker-types` to wrangler generated types (as recommended by Cloudflare [1]/[2])
- Remove unnecessary scripts in `api/package.json`
- Add `types:check` script in `api/package.json`
- Change `api/package.json` `name` to `@vetro/api` to match other packages
- Set `"private": true` in `api/package.json` to prevent publishing

### Related issue(s)

Closes #
Related to #

### Checklist

- [x] Manual testing passed.

[1]: https://developers.cloudflare.com/workers/best-practices/workers-best-practices/#generate-binding-types-with-wrangler-types
[2]: https://developers.cloudflare.com/workers/languages/typescript/#migrating


api/wrangler.jsonc
```diff
@@ -17,14 +22,27 @@
     "staging": {
       "name": "vetro-api-staging",
       "vars": {
+        "MERKL_OPPORTUNITY_ID": "",
+        "ORIGINS": "http://localhost:5173,https://vetro.letshamsterdance.xyz",
```
gabmontes: This repository may go public soon. We should not commit any staging URL.

gabmontes (changes requested):

api/wrangler.jsonc
gndelia: note that if we go public, the PRs will be public and the PR previews (which point to staging url) will also be public..

joshuasing: I have moved it to Cloudflare, but agreed ^ - They're protected by Cloudflare Access, so they won't be accessible. Though, if this repository is open source, does it matter if they are accessible?

gabmontes: @joshuasing I don't think we want to leak the staging domain. But if you and @jcvernaleo are ok with that, I will not oppose.

gabmontes (approved):
LGTM

gndelia (approved):

PR #93, opened 2026-02-17
Add delete exit ticket feature

ArturDolzan:
### Description

This PR implements the delete exit ticket. It includes Modal base component, danger button variant, useCancelWithdraw hook with optimistic cache updates, cancelled ticket status, and i18n translations (en/es).

### Screenshots

https://github.com/user-attachments/assets/408cfd70-c3ba-4827-8994-00ff19967991

### Related issue(s)

Related to #23 

### Checklist

- [x] Manual testing passed.


web/src/pages/earn/components/exitTickets/actionsCell.tsx
```diff
@@ -13,21 +17,46 @@ type Props = {
...
+        {status === "ready" && (
+          <Button size="xSmall" variant="primary">
+            {t("pages.earn.exit-tickets.withdraw")}
+          </Button>
+        )}
+        <Tooltip content={t("pages.earn.exit-tickets.delete-tooltip")}>
+          <ButtonIcon
+            onClick={() => setIsModalOpen(true)}
+            size="xSmall"
+            variant="secondary"
```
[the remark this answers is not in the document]
ArturDolzan: Applied here 01748a7155247e7496e75f559166d8e0ab732482.

web/src/pages/earn/components/exitTickets/deleteTicketModal.tsx
```diff
@@ -0,0 +1,84 @@
...
+            variant="secondary"
+          >
+            {t("pages.earn.exit-tickets.delete-btn-cancel")}
+          </Button>
+          {isConnected ? (
+            <Button
+              disabled={isDeleting}
+              onClick={handleDelete}
+              size="xSmall"
+              variant="danger"
```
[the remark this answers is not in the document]
ArturDolzan: Applied here 01748a7155247e7496e75f559166d8e0ab732482.

gndelia (commented):
In the video, I see that after the cancel is successful, both the Earned amount and the Staked position increment by the same amount that it was canceled (10 units of token), but the staked position is TestUSD and the other one is SVUSD. Is it correct that both values update?

Edit: It may not be wrong. But I guess my confusion comes from the fact that "Earned" contains the deposit itself 🤔

web/src/components/base/modal/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }, []);
+
+  function handleTransitionEnd(e: TransitionEvent) {
+    if (e.target !== e.currentTarget) {
+      return;
+    }
+    if (!isOpen) {
+      onClose();
+    }
+  }
```
gndelia: I think we're using this in several parts - perhaps we could extract it into a reusable hook?

web/src/hooks/useCancelWithdraw.ts
```diff
@@ -0,0 +1,123 @@
...
+        onStatusChange?.("cancelling");
+      });
+
+      emitter.on("user-signing-cancel-withdraw-error", function () {
+        onStatusChange?.("failed");
+      });
+
+      emitter.on(
+        "cancel-withdraw-transaction-reverted",
+        function (receipt: TransactionReceipt) {
```
gndelia: A minor detail, but I think you should be able to remove these typings and the compiler will automatically infer it's a `TransactionReceipt`. Same on line  84

```suggestion
        function (receipt) {
```

ArturDolzan:
> In the video, I see that after the cancel is successful, both the Earned amount and the Staked position increment by the same amount that it was canceled (10 units of token), but the staked position is TestUSD and the other one is SVUSD. Is it correct that both values update?

As far as I understand, both values updating is correct. The Earned amount is the user's `svUSD` (shares) balance, and the Staked position is the equivalent value in `vUSD` (assets) derived via convertToAssets(shares).

gndelia:
> But I guess my confusion comes from the fact that "Earned" contains the deposit itself 🤔

I see. Then I think it's ok.

Not a blocker of this PR, but I think also then the title of those labels may be incorrect. The "Earned amount" should include earnings, but as is, it's including the deposit + earnings (That's why it's matching the staked position). Splitting balance from earning may be difficult, the easier change could be just changing the title so something more accurate. But we can discuss it with @dvnahuel  - not something from this PR

web/src/pages/earn/components/exitTickets/deleteTicketModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+        setIsDeleting(true);
+      }
+      if (status === "failed") {
+        setIsDeleting(false);
+      }
+    },
+    onSuccess() {
+      onClose();
+      onSuccess();
+    },
```
gndelia: `onSuccess` calling `onSuccess`

Image

😆

gndelia: So here it seems we have 2 functions that do something very similar. `onStatusChange` handles the status changes, and `onSuccess` also handles a status change: The one about being succesful. It seems we can merge these 2 functions into one 🤔  What do you think?

web/src/hooks/useCancelWithdraw.ts
```diff
@@ -0,0 +1,123 @@
...
+import { useEthereumWalletClient } from "./useEthereumWalletClient";
+import { useMainnet } from "./useMainnet";
+import { stakedBalanceQueryKey } from "./useStakedBalance";
+
+type CancelWithdrawStatus = "cancelling" | "completed" | "failed";
+
+type Params = {
+  assets: bigint;
+  onStatusChange?: (status: CancelWithdrawStatus) => void;
+  onSuccess?: VoidFunction;
```
gndelia: I like the flexibility and all that, but the only callers of these hooks are us, and we know for sure we're going to pass these functions. So let's make them mandatory, so we don't need to add `?.` that are really not needed

We can always make them optional if needed

web/src/pages/earn/components/exitTickets/deleteTicketModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return (
+    <Modal onClose={onClose}>
+      <div className="flex w-[448px] flex-col gap-6 rounded-lg bg-white p-6 shadow-xl">
+        <div className="flex flex-col gap-2">
+          <h4 className="text-base font-semibold tracking-tight text-gray-900">
+            {t("pages.earn.exit-tickets.delete-title")}
+          </h4>
+          <p className="text-xsm leading-5 font-normal text-gray-500">
```
gndelia: I think `text-xsm` already includes `leading-5`

Image

```suggestion
          <p className="text-xsm font-normal text-gray-500">
```

gndelia (commented):
Sounds like the modal stories are missing 🤔

web/src/components/base/modal/index.tsx
ArturDolzan: Good point. Refactored here dbceeb4417afe81057bb73e6498c5aff867869ad.

web/src/hooks/useCancelWithdraw.ts
ArturDolzan: Applied here dbceeb4417afe81057bb73e6498c5aff867869ad.

web/src/pages/earn/components/exitTickets/deleteTicketModal.tsx
ArturDolzan: Good point. Applied here dbceeb4417afe81057bb73e6498c5aff867869ad.

web/src/hooks/useCancelWithdraw.ts
ArturDolzan: Agreed. Applied here dbceeb4417afe81057bb73e6498c5aff867869ad.

web/src/pages/earn/components/exitTickets/deleteTicketModal.tsx
ArturDolzan: Good catch! Fixed here dbceeb4417afe81057bb73e6498c5aff867869ad.

ArturDolzan:
> Sounds like the modal stories are missing 🤔

Added here cbd9c84b343adb03c8ea7b3883b05d505c1499ff. Thanks!


https://github.com/user-attachments/assets/0db08487-63db-4ee5-89b2-60721741e5cf

gndelia (approved):

dvnahuel:
One minor UI point of improvement the 'actions' colum doesn't seem to be aligned with the other elements of the page.
Screenshot 2026-02-17 at 11 27 52 AM

PR #106, opened 2026-02-19
Add instant withdrawal for whitelisted addresses

ArturDolzan:
### Description

This PR implements the whitelisted addresses (or all users when cooldown is disabled) can now withdraw directly via the ERC-4626 withdraw function, bypassing the 7-day cooldown request flow.

- Add instantWithdrawWhitelist to StakingVault ABI
- Add getInstantWithdrawWhitelist public action with tests
- Add useCanInstantWithdraw hook mirroring contract logic
- Add useInstantWithdraw hook using viem-erc4626 withdraw
- Conditionally show 1-step vs 3-step flow in withdraw form
- Hide exit tickets section when instant withdraw is available

### Screenshots

**When disabled for everyone**

https://github.com/user-attachments/assets/a8eb7659-384e-4b37-8ca3-e996a5da1eee

**When user is whitelisted**

https://github.com/user-attachments/assets/cfee91f9-9e85-4d93-83ab-e5d5a421e283

### Related issue(s)

Related to #23 

### Checklist

- [x] Manual testing passed.


dvnahuel:
LGTM!

web/src/pages/earn/hooks/useCanInstantWithdraw.ts
```diff
@@ -0,0 +1,41 @@
...
+    queryKey: [
+      "instant-withdraw-whitelist",
+      chain.id,
+      stakingVaultAddress,
+      account,
+    ],
+  });
+
+  // Mirror contract: !cooldownEnabled || isWhitelisted
+  return cooldownEnabled === false || isWhitelisted === true;
```
gndelia: Note that this causes a wrong result while loading. You may want to move this inside the fetcher so you get undefined while loading

web/src/hooks/useInstantWithdraw.ts
```diff
@@ -0,0 +1,110 @@
...
+      // useStakedBalance uses ensureQueryData to read shares from cache
+      await queryClient.invalidateQueries({
+        queryKey: sharesBalanceKey,
+      });
+
+      queryClient.invalidateQueries({
+        queryKey: stakedKey,
+      });
+
+      queryClient.invalidateQueries({
```
gndelia: Do we need to invalidate this one? I think exit tickets are not for instant withdrawals

web/src/hooks/useInstantWithdraw.ts
```diff
@@ -0,0 +1,110 @@
...
+        hash,
+      });
+
+      if (receipt.status === "reverted") {
+        updateNativeBalanceAfterReceipt(receipt);
+        onStatusChange?.("failed");
+        return;
+      }
+
+      updateNativeBalanceAfterReceipt(receipt);
```
gndelia: We can move this call before the if, as it should be called after waiting for a receipt, regardless of the outcome
This way we avoid the duplicated call

web/src/hooks/useInstantWithdraw.ts
```diff
@@ -0,0 +1,110 @@
...
+      onStatusChange?.("completed");
+      onSuccess?.();
+
+      // Optimistically update staked balance
+      queryClient.setQueryData(stakedKey, (old: bigint | undefined) =>
+        old !== undefined ? old - assets : old,
+      );
+    },
```
gndelia, line 91: Should we optimistically update vusd balance here, from the tokens withdraw? It seems so

web/src/pages/earn/hooks/useCanInstantWithdraw.ts
ArturDolzan: Applied here 537b6ae03f196aeb5081cd7318981da70f748f79. thanks!

web/src/hooks/useInstantWithdraw.ts
ArturDolzan: Applied here 537b6ae03f196aeb5081cd7318981da70f748f79. thanks!

web/src/hooks/useInstantWithdraw.ts
ArturDolzan: Applied here 537b6ae03f196aeb5081cd7318981da70f748f79. thanks!

web/src/hooks/useInstantWithdraw.ts
ArturDolzan: Applied here 537b6ae03f196aeb5081cd7318981da70f748f79. thanks!

gndelia (approved):

Issue #134, opened 2026-03-04
Implement Borrow page


gndelia:

gndelia:
Accidentally closed

PR #135, opened 2026-03-04
Borrow: Part II

gndelia:
### Description

This PR adds the "Markets table" that list markets to borrow from.

Note that the markets shown are testing ones, and they need to eventually be replaced with the appropiate marketId.

### Screenshots

image

image

image

image

### Related issue(s)

Related to #134

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/components/base/displayAmount.tsx
+++ b/web/src/components/base/displayAmount.tsx
@@ -1,0 +1,70 @@
+import { TokenLogo } from "components/tokenLogo";
+import { Tooltip } from "components/tooltip";
+import { type ComponentType, Fragment, type ReactNode } from "react";
+import { type Token } from "types";
+import { formatNumber } from "utils/format";
+import { formatUnits, parseUnits } from "viem/utils";
+
+type CustomContainer = ComponentType<{
+  children?: ReactNode;
+}>;
+
+const DefaultTextContainer: CustomContainer = ({ children }) => (
+  <span>{children}</span>
+);
+
+type Props = {
+  amount: bigint;
+  amountContainer?: CustomContainer;
+  container?: CustomContainer;
+  symbolContainer?: CustomContainer;
+  showSymbol?: boolean;
+  showTokenLogo?: boolean;
+  token: Token;
+};
+
+export const DisplayAmount = function ({
+  amount,
+  amountContainer: AmountContainer = DefaultTextContainer,
+  container: Container = Fragment,
+  showSymbol = true,
+  showTokenLogo = true,
+  symbolContainer: SymbolContainer = DefaultTextContainer,
+  token,
+}: Props) {
+  const stringAmount = formatUnits(amount, token.decimals);
+  const formattedAmount = formatNumber(stringAmount);
+
+  const one = parseUnits("1", token.decimals);
+
+  const notZero = amount !== 0n;
+  // Only show dots for small numbers (x < 1), if the amount is not zero and we're truncating decimals
+  const showDots = amount < one && notZero && formattedAmount !== stringAmount;
+  return (
+    <Tooltip
+      content={
+        notZero ? (
+          <span className="flex items-center gap-x-1">
+            {showTokenLogo && <TokenLogo {...token} />}
+            <span>{`${new Intl.NumberFormat("en-US", {
+              maximumFractionDigits: token.decimals,
+              useGrouping: true,
+            }).format(
+              // @ts-expect-error NumberFormat.format accept strings, typings are wrong. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#parameters
+              stringAmount,
+            )} ${token.symbol}`}</span>
+          </span>
+        ) : null
+      }
+    >
+      <Container>
+        <AmountContainer>{`${formattedAmount}${
+          showDots ? "..." : ""
+        }`}</AmountContainer>
+        {showSymbol ? (
+          <SymbolContainer>{` ${token.symbol}`}</SymbolContainer>
+        ) : null}
+      </Container>
+    </Tooltip>
+  );
+};
--- a/web/src/components/base/fiatValue.tsx
+++ b/web/src/components/base/fiatValue.tsx
@@ -1,0 +1,58 @@
+import { type QueryStatus } from "@tanstack/react-query";
+import { useTokenPrices } from "hooks/useTokenPrices";
+import { type ComponentProps } from "react";
+import Skeleton from "react-loading-skeleton";
+import type { Token } from "types";
+import { formatFiatNumber } from "utils/format";
+import { getTokenPrice } from "utils/token";
+import { formatUnits } from "viem";
+
+import { ErrorBoundary } from "./errorBoundary";
+
+const RenderFiatValueUnsafe = function ({
+  customFormatter = formatFiatNumber,
+  queryStatus,
+  token,
+  value,
+}: {
+  customFormatter?: (amount: string) => string;
+  queryStatus: QueryStatus;
+  token: Token;
+  value: bigint | undefined;
+}) {
+  const { data: pricesData, status: pricesStatus } = useTokenPrices({
+    retryOnMount: false,
+  });
+
+  if (value !== undefined && pricesData !== undefined) {
+    const stringBalance = formatUnits(value, token.decimals);
+    const price = getTokenPrice(token, pricesData);
+
+    return (
+      <>
+        {customFormatter(
+          (parseFloat(stringBalance) * parseFloat(price)).toFixed(
+            token.decimals,
+          ),
+        )}
+      </>
+    );
+  }
+
+  // Check for errors from either source
+  if (queryStatus === "error" || pricesStatus === "error") {
+    return <>-</>;
+  }
+
+  // Loading state (either balance or prices are loading)
+  return <Skeleton className="h-full" containerClassName="w-8" />;
+};
+
+export const RenderFiatValue = (
+  props: ComponentProps<typeof RenderFiatValueUnsafe>,
+) => (
+  // Prevent crashing if a price is missing or wrongly mapped
+  <ErrorBoundary fallback="-">
+    <RenderFiatValueUnsafe {...props} />
+  </ErrorBoundary>
+);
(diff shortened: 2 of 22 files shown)


web/src/utils/token.ts
```diff
@@ -1,5 +1,17 @@
...
 import { formatUnits, parseUnits as viemParseUnits } from "viem";
+
+export const getTokenPrice = function (
+  token: Token,
+  prices: Record<string, string> | undefined,
+) {
+  const priceSymbol = (
+    token.extensions?.priceSymbol ?? token.symbol
+  ).toUpperCase();
+  const price = prices?.[priceSymbol] ?? "0";
```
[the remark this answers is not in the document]
gndelia: 0 is ok

web/src/components/borrow/collateralCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+type Props = {
+  marketId: Hash;
+  token: Token;
+};
+
+export function CollateralCell({ marketId, token }: Props) {
+  const { data: value, status } = useMarketCollateral(marketId);
+
+  return (
+    <div className="items-end-end flex flex-col gap-y-0.5">
```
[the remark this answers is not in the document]
gndelia: Done in [b1bb8f3](https://github.com/vetro-protocol/vetro-monorepo/pull/135/commits/b1bb8f3a8cd225fe5c45ff5acf24bc8ae39c7e8c)

web/src/hooks/borrow/useMarketCollateral.ts
```diff
@@ -0,0 +1,28 @@
...
+              }
+            }
+          }`,
+          variables: { marketId },
+        }),
+        headers: { "Content-Type": "application/json" },
+        method: "POST",
+      });
+      const json = await response.json();
+      return BigInt(json.data.marketByUniqueKey.state.collateralAssets);
```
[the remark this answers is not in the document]
gndelia: The current error handling is sufficient — `response.json()` will throw if the body is not valid JSON, and accessing nested properties will throw if the shape is wrong. React Query catches both cases and surfaces the error via `status: "error"`, which the component already handles. Adding `response.ok` is reasonable but the full 30+ line validation suggested is over-engineered for this use case. React Query also retries failed queries automatically.

web/src/components/borrow/marketsTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+        id: "loan",
+        meta: { className: "grow", width: "143px" },
+      },
+      {
+        cell: ({ row }) => (
+          <CollateralCell
+            marketId={row.original.marketId}
+            token={row.original.collateralToken}
+          />
+        ),
```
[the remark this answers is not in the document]
gndelia: This is a premature optimization — the markets list is small (testing markets), and each query is cached by React Query so re-renders do not re-fetch. Batching would require a different GraphQL query shape, adding complexity for no current benefit. We can revisit if the list grows significantly.

web/src/components/base/displayAmount.tsx
```diff
@@ -0,0 +1,70 @@
...
+        notZero ? (
+          <span className="flex items-center gap-x-1">
+            {showTokenLogo && <TokenLogo {...token} />}
+            <span>{`${new Intl.NumberFormat("en-US", {
+              maximumFractionDigits: token.decimals,
+              useGrouping: true,
+            }).format(
+              // @ts-expect-error NumberFormat.format accept strings, typings are wrong. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#parameters
+              stringAmount,
+            )} ${token.symbol}`}</span>
```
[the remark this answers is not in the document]
gndelia: The `@ts-expect-error` is intentional and correct. TypeScript types `Intl.NumberFormat.format` as `(value?: number | bigint) => string`, so passing a string IS a type error. However, the MDN docs confirm strings are accepted at runtime (coerced to numeric), and the test verifies this: `"123456789.123456789012345678"` formats to `"123,456,789.123456789012345678"` preserving full precision. Using `number` would actually LOSE precision for large values due to IEEE 754 limits — the opposite of what we want.

web/src/components/borrow/marketsTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+              viewBox="0 0 16 16"
+              width="16"
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              <path
+                d="M6.22 4.22a.75.75 0 0 1 1.06 0l3.5 3.5a.75.75 0 0 1 0 1.06l-3.5 3.5a.75.75 0 0 1-1.06-1.06L9.19 8 6.22 5.28a.75.75 0 0 1 0-1.06Z"
+                fill="currentColor"
+              />
+            </svg>
+          </ButtonIcon>
```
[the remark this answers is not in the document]
gndelia: Good catch — fixed by adding `aria-label` with a translated string.

web/src/components/base/fiatValue.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (value !== undefined && pricesData !== undefined) {
+    const stringBalance = formatUnits(value, token.decimals);
+    const price = getTokenPrice(token, pricesData);
+
+    return (
+      <>
+        {customFormatter(
+          (parseFloat(stringBalance) * parseFloat(price)).toFixed(
+            token.decimals,
+          ),
```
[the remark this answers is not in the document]
gndelia: This is a fiat value display (USD). Precision beyond ~15 significant digits (IEEE 754 limit) is irrelevant for dollar amounts. The value is already formatted to fiat-appropriate decimals via `formatFiatNumber`. Adding a big decimal library for approximate fiat display would be over-engineering.

web/src/components/borrow/marketsTable.tsx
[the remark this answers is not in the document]
gndelia: Done in [85bd5ae](https://github.com/vetro-protocol/vetro-monorepo/pull/135/commits/85bd5aec9d1248c93a75e0ae96b74a6be99ce864)

web/src/components/borrow/marketsTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+              value={
+                row.original.totalSupplyAssets - row.original.totalBorrowAssets
+              }
+            />
+          </span>
+        ),
+        header: () => <Header text={t("pages.borrow.available-to-borrow")} />,
+        id: "available",
+        meta: { className: "justify-end", width: "120px" },
+      },
```
gabmontes, line 78: The "Pool size" and "Available to borrow" numbers do not have a ticker. That's confusing to me. cc/ @dvnahuel

web/src/hooks/borrow/useMarketCollateral.ts
```diff
@@ -0,0 +1,28 @@
...
+                collateralAssets
+              }
+            }
+          }`,
+          variables: { marketId },
+        }),
+        headers: { "Content-Type": "application/json" },
+        method: "POST",
+      });
+      const json = await response.json();
```
gabmontes, line 21: What about using something like `runQuery()` from https://github.com/vetro-protocol/vetro-monorepo/blob/master/api/src/subgraph.ts?

web/src/i18n/locales/es/translation.json
```diff
@@ -14,7 +14,21 @@
       "title": "Hola Analíticas"
     },
     "borrow": {
-      "title": "Hola Préstamos"
+      "available-to-borrow": "Disponible para pedir prestado",
+      "borrow-apr": "TAE de préstamo",
```
gabmontes: In Argentina we use to say "TEA" instead of "TAE".

web/src/i18n/locales/es/translation.json
```diff
@@ -14,7 +14,21 @@
...
+      "collateral": "Colateral",
+      "health-factor": "Factor de salud",
+      "liquidation-price": "Precio de liquidación",
+      "loan": "Préstamo",
+      "ltv": "LTV",
+      "manage": "Gestionar",
+      "market-details": "Detalles del mercado",
+      "markets-title": "Mercados",
+      "no-positions-description": "Deposite colateral para pedir prestado {{symbol}} y poner sus activos a trabajar.",
```
gabmontes, line 27: I think "colateral" does not translate to "colateral". Let's check this.

web/src/i18n/locales/es/translation.json
```diff
@@ -14,7 +14,21 @@
...
+      "no-positions-title": "Aún no hay posiciones de préstamos activos",
+      "pool-size": "Tamaño del pool",
```
gabmontes, line 29: Didn't we used to translate "pool" to "fondo" or something similar for Vesper?

web/src/components/borrow/marketsTable.tsx
gndelia: I guess they should use the $ as they are in USD

image

the ones showing the token do use the token ticker

web/src/hooks/borrow/useMarketCollateral.ts
gndelia: I went for the simplest as (hopefully) this is the only value we need to get from Morpho's GraphQL. We can move it to the API BE if you prefer, though

web/src/i18n/locales/es/translation.json
gndelia: We do use colateral in spanish

web/src/components/borrow/marketsTable.tsx
dvnahuel: Yep, we should add the ($) to the above one.

gndelia: Added in [c49d073](https://github.com/vetro-protocol/vetro-monorepo/pull/135/commits/c49d073476be4fa1e4a08676f39af7b3003c5e04)

image

web/src/i18n/locales/es/translation.json
gndelia: Updated in [c771884](https://github.com/vetro-protocol/vetro-monorepo/pull/135/commits/c77188442fdf1bc3bf8737896973554fee507785)

web/src/i18n/locales/es/translation.json
gndelia: I think we did. I guess in the way of using some words as-is with no translation, I forgot that pool could be translated to "Fondo". However, changing this will require changing past translations. Let me do so in a separated issue https://github.com/vetro-protocol/vetro-monorepo/issues/136

web/src/hooks/borrow/useMarketCollateral.ts
gabmontes: No need to move it to the backend.

gabmontes (approved):

PR #181, opened 2026-03-11
Update contract addresses

gndelia:
### Description

Updating the contract addresses based on the new deployments [here](https://github.com/hemilabs/vetro-contracts/pull/98)


gabmontes (approved):

ArturDolzan (approved):
Do we need to change anything in the `setup.ts` script?

gndelia:
> Do we need to change anything in the `setup.ts` script?

nope, the setup script reads the symbols from the token list, and gets the list of addresses from there. VUSD symbol was already listed

Issue #145, opened 2026-03-05
Implement analytics page


ArturDolzan:

ArturDolzan:
@gabmontes I think this issue can be closed. Is there anything pending for this page?

gabmontes:
Yeah. Let's close it and open individual issues for any addition or improvement.

PR #215, opened 2026-03-17
Add Analytics page with allocation cards

ArturDolzan:
### Description

This PR enables the Analytics route in the navbar and implement the page UI with two side-by-side allocation cards (TVL and Yield).

Each card shows a proportional horizontal bar chart and a color-coded legend. Data is mocked to match the shape of the `/analytics/treasury` API response, to be replaced by hooks in a follow-up PR.

Also adds Storybook stories for AllocationCard, AllocationChart and AllocationLegend.

### Screenshots

Storybook:

https://github.com/user-attachments/assets/a09d2383-1930-42bc-963c-69b624ff3a89

App:

https://github.com/user-attachments/assets/08ab1993-2d62-4424-8f84-f3cba0591c5c

### Related issue(s)

Related to #145 

### Checklist

- [x] Manual testing passed.


web/src/pages/analytics/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  { amount: 37_500_000, color: "bg-amber-400", label: "DAI" },
+  { amount: 12_500_000, color: "bg-rose-400", label: "HemiBTC CDPs" },
+];
+
+const yieldItems: AllocationItem[] = [
+  { amount: 147, color: "bg-emerald-400", label: "Morpho" },
+  { amount: 80, color: "bg-blue-400", label: "Aave" },
+  { amount: 100, color: "bg-rose-400", label: "Summer.fi" },
+  { amount: 64, color: "bg-amber-400", label: "Compound" },
+];
```
gabmontes: Let's think about how to do this in the final implementation.

gabmontes (dismissed):

dvnahuel:
Nice! I'd also suggest adding this interaction: hovering over a detail item or one of the bars in the chart would dim the others, so you can clearly see which bar connects to which item. 

Designs [here](https://www.figma.com/design/5qiF7MSvF4xfzQiS11ZiBk/Vetro--vUSD-?node-id=1703-2253&t=ThJdE1avWwYprPWN-1)

Screenshot:

Screenshot 2026-03-17 at 11 51 26 AM

ArturDolzan:
> Nice! I'd also suggest adding this interaction: hovering over a detail item or one of the bars in the chart would dim the others, so you can clearly see which bar connects to which item.
> 
> Designs [here](https://www.figma.com/design/5qiF7MSvF4xfzQiS11ZiBk/Vetro--vUSD-?node-id=1703-2253&t=ThJdE1avWwYprPWN-1)
> 
> Screenshot:
> 
> Screenshot 2026-03-17 at 11 51 26 AM

@dvnahuel ohh I totally forgot that. Added on fd4449f0df02f5e2654c0a512487c734586ab859. Thanks!

https://github.com/user-attachments/assets/4f94cd1e-8909-49a8-a6ac-f653169874a8

gabmontes (dismissed):

web/src/pages/analytics/index.tsx
gndelia: I would suggest moving this into a useQuery, returning this data as part of the queryFn (even with a setTimeout of 2000ms for example)

This way you can already code the loading and error states, and once the API is completed, you just switch the mocked data with the call and everything would work. Thoughts? Or would you prefer doing so in a separated PR ?

dvnahuel:
> > Nice! I'd also suggest adding this interaction: hovering over a detail item or one of the bars in the chart would dim the others, so you can clearly see which bar connects to which item.
> > Designs [here](https://www.figma.com/design/5qiF7MSvF4xfzQiS11ZiBk/Vetro--vUSD-?node-id=1703-2253&t=ThJdE1avWwYprPWN-1)
> > Screenshot:
> > Screenshot 2026-03-17 at 11 51 26 AM
> 
> @dvnahuel ohh I totally forgot that. Added on [fd4449f](https://github.com/vetro-protocol/vetro-monorepo/commit/fd4449f0df02f5e2654c0a512487c734586ab859). Thanks!
> 
>  Gravacao.de.Tela.2026-03-17.as.16.08.22.mov

Looks great, thanks!

web/src/index.css
```diff
@@ -27,6 +27,8 @@
   --line-height-xsm: 1.25rem; /* 20px */
   --line-height-mid: 1.375rem; /* 22px */
 
+  --radius-xs: 1px;
```
gndelia: this is updating tailwind's default theme (in which `xs` is 2px). Could this break somewhere?

Perhaps we could do what tailwind v3 used for `1px`, which was using `-px`

That would be


```suggestion
  --radius-px: 1px;
```

web/src/pages/analytics/index.tsx
ArturDolzan: I thought for upcoming requests, but makes sense including those behaviors in this PR as well. Applied on 8035e93e0fab5879d8047675aa9bfbd618ccdf66. Thanks!

web/src/index.css
ArturDolzan: Good catch! Fixed on 8035e93e0fab5879d8047675aa9bfbd618ccdf66. Thanks!

web/src/pages/analytics/index.tsx
ArturDolzan: I've updated the videos in the description of this issue to reflect this.

web/src/pages/analytics/components/allocationCard/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <div className="flex flex-1 flex-col">
+      <div className="px-6 md:px-14">
+        <div className="flex flex-col gap-3 border-t border-blue-500 py-6">
+          <div className="flex items-center justify-between">
+            <span className="text-b-medium text-gray-900">{label}</span>
+            <div className="size-4">{icon}</div>
+          </div>
+          <h3 className="text-h3 font-semibold text-gray-900">
```
gndelia: `text-h3` already includes `font-semibold`

```suggestion
          <h3 className="text-h3 text-gray-900">
```

See [text-h3 definition](https://github.com/vetro-protocol/vetro-monorepo/blob/d9e363729e839b23f10c13856c5b1a132bf66af8/web/src/index.css#L114)

ArturDolzan: Applied on cb60d1f575aa1fd74beb67514a56470f213c71b8.

gndelia (dismissed):

web/src/hooks/useAnalyticsTreasury.ts
```diff
@@ -0,0 +1,31 @@
...
+// value: total count of active strategies across all tokens
+const mockData: AnalyticsTreasury = {
+  items: [
+    { amount: 147, color: "bg-emerald-400", label: "Morpho" },
+    { amount: 80, color: "bg-blue-400", label: "Aave" },
+    { amount: 100, color: "bg-rose-400", label: "Summer.fi" },
+    { amount: 64, color: "bg-amber-400", label: "Compound" },
+  ],
+  value: "8 protocols",
+};
```
gabmontes: The mock data should be set to match what the API returns. In this case, it also contains colors and labels, mixing data and display responsibilities. Processing of the data returned by the API into something the UI component can use to be rendered does not belong here.

web/src/hooks/useAnalyticsTvl.ts
```diff
@@ -0,0 +1,34 @@
...
+  items: [
+    { amount: 37_500_000, color: "bg-blue-400", label: "USDT" },
+    { amount: 37_500_000, color: "bg-emerald-400", label: "USDC" },
+    { amount: 37_500_000, color: "bg-amber-400", label: "DAI" },
+    { amount: 12_500_000, color: "bg-rose-400", label: "HemiBTC CDPs" },
+  ],
+  value: "$125,000,000.00",
+};
+
+export const useAnalyticsTvl = () =>
```
gabmontes: I don't think there is such thing as "analytics TVL". The charts are fed only from /analytics/treasury. The API does provide two more endpoints relevant to the analytics page but not for these cards. Ones is /analytics/totals and the other is /variable-stake/apy. Both will feed the data boxes at the top of the allocation cards. This hook should not exist. The mocked data should be also derived from the useAnalyticsTreasury hook.

gabmontes (changes requested):

web/src/hooks/useAnalyticsTreasury.ts
ArturDolzan: Agreed. Added on bdd612ceaff2d9acfc0bf785ea885afe078f5c08. Thanks for pointing this out!

web/src/hooks/useAnalyticsTvl.ts
ArturDolzan: Yes, I misunderstood that 😢 Fixed on bdd612ceaff2d9acfc0bf785ea885afe078f5c08.

gndelia (approved):

gabmontes (approved):
Not sure about the calculations but let's move ahead. When we have actual numbers, we can validate the page renders properly.

Issue #248, opened 2026-03-20
Show the size of the sVUSD withdrawal queue


gabmontes:
### Feature or change description

Show in the analytics page how much $ is locked in the sVUSD withdrawal queue.

### Proposed implementation

All the active exit tickets can be obtained by querying the subgraph as follows:

```gql
query {
  exitTickets(where: { cancelTxHash: null, claimTxHash: null }) {
    shares
  }
}
```

That query, and proper pagination handling, should be done in the API, summarized, then exposed in i.e. a `variable-stake/withdrawal-queue` endpoint. 

Finally, that shall be show in in the analytics page.

gabmontes:
The route `GET /variable-stake/exit-queue` was added in #249. The data needs to be shown in the analytics page now.

@dvnahuel please provide a UI design.

dvnahuel:
Just to double check this is related to the Earn page and not to the swap one, right? @gabmontes @ArturDolzan

dvnahuel:
Designs [here](https://www.figma.com/design/5qiF7MSvF4xfzQiS11ZiBk/Vetro--vUSD-?node-id=1904-2474&t=vKqt2MiYXb6Uq1Ee-1)

gabmontes:
Analytics page. Yes!

PR #249, opened 2026-03-21
Exit queue summary data

gabmontes:
### Description

Update the subgraph to create and maintain an entity that holds the size and tokens in the sVUSD exit queue. Update the API to expose that information to the app.

### Related issue(s)

Related to #248

### Checklist

- [x] Manual testing passed.


api/src/variable-stake.ts
```diff
@@ -141,3 +142,30 @@ export async function getUserExitTickets({
...
+  if (!exitTicketQueueSummary) {
+    return {
+      openTickets: 0,
+      shares: 0n,
+    };
+  }
+  return {
+    openTickets: exitTicketQueueSummary.openTickets,
+    shares: BigInt(exitTicketQueueSummary.shares),
+  };
```
gndelia, line 170: `bigint` can't be serialized over the network

```suggestion
  if (!exitTicketQueueSummary) {
    return {
      openTickets: 0,
      shares: "0",
    };
  }
  return {
    openTickets: exitTicketQueueSummary.openTickets,
    shares: exitTicketQueueSummary.shares,
  };
```

So this will either throw, or being converted to string automatically, which means it does nothing converting these to `bigint`

gabmontes: Yup. That's why `convertBigIntsToString` is called before returning the data.

https://github.com/vetro-protocol/vetro-monorepo/pull/249/changes#diff-c7daeed68470fce500b64961dc1986f90d3c903d7679f553701de7aed7aa72a6R170

gndelia (approved):

ArturDolzan (approved):

PR #253, opened 2026-03-23
Fix the activity list bottom blur

### LINKED ISSUE #227: Last item in the Drawer activity is blurred
### Description

See image 

Image

The bottom part is blurred (ok) to allow scrolling down, but the last item of the whole list (After completely scrolling down) is blurred, but it shouldn't be.

### Steps to reproduce

Make several operations so the Activity list has a scroll. When scrolling is available, the bottom appears blurred to indicate there are operations to scroll down to.

Once the user has scrolled to the bottom, the blur should dissapear, but it's there

gabmontes:
### Description

This PR solves the blur at the bottom of the activity list by 1) removing the bottom padding of the list container and 2) adding a bottom padding to the list itself. It also reduces the height of the blur overlay to match the empty space at the bottom of the list (6 bottom list padding + 3 bottom list item padding).

### Screenshots

https://github.com/user-attachments/assets/28bd4659-3b9f-45d4-b3f1-06820a180461

### Related issue(s)

Closes #227

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/components/base/activityList/index.tsx
+++ b/web/src/components/base/activityList/index.tsx
@@ -3,18 +3,20 @@ import { ActivityListEmptyState } from "./activityListEmptyState";
 import type { Activity } from "./types";
 
 type Props = {
+  className?: string;
   hasTransactions: boolean;
   items: (Activity & { href?: string })[];
   onResetFilters: VoidFunction;
 };
 
 export const ActivityList = ({
+  className,
   hasTransactions,
   items,
   onResetFilters,
 }: Props) =>
   items.length > 0 ? (
-    <div className="flex flex-col pb-6">
+    <div className={`flex flex-col ${className}`}>
       {items.map((item) => (
         <ActivityItem key={item.txHash} {...item} />
       ))}
--- a/web/src/components/base/activityList/index.tsx
+++ b/web/src/components/base/activityList/index.tsx
@@ -3,24 +3,27 @@ import { ActivityListEmptyState } from "./activityListEmptyState";
 import type { Activity } from "./types";
 
 type Props = {
-  className?: string;
   hasTransactions: boolean;
   items: (Activity & { href?: string })[];
   onResetFilters: VoidFunction;
 };
 
 export const ActivityList = ({
-  className,
   hasTransactions,
   items,
   onResetFilters,
 }: Props) =>
   items.length > 0 ? (
-    <div className={`flex flex-col ${className}`}>
-      {items.map((item) => (
-        <ActivityItem key={item.txHash} {...item} />
-      ))}
-    </div>
+    <>
+      <div className="absolute inset-0 overflow-y-auto px-4 pt-6 md:px-6">
+        <div className="flex flex-col pb-6">
+          {items.map((item) => (
+            <ActivityItem key={item.txHash} {...item} />
+          ))}
+        </div>
+      </div>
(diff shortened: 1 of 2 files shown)


web/src/components/base/activityList/index.tsx
```diff
@@ -14,7 +14,7 @@ export const ActivityList = ({
   onResetFilters,
 }: Props) =>
   items.length > 0 ? (
-    <div className="flex flex-col">
+    <div className="flex flex-col pb-6">
```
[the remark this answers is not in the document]
gabmontes: Ok.

web/src/components/walletDrawer/walletDrawerContent.tsx
```diff
@@ -139,14 +139,14 @@ export function WalletDrawerContent() {
...
-          <div className="absolute inset-0 overflow-y-auto px-4 py-6 md:px-6">
+          <div className="absolute inset-0 overflow-y-auto px-4 pt-6 md:px-6">
             <ActivityList
               hasTransactions={activities.length > 0}
               items={itemsWithHref}
               onResetFilters={() => setSelectedFilters(allFilters)}
             />
           </div>
-          <div className="pointer-events-none absolute inset-x-0 bottom-0 h-34 bg-gradient-to-b from-transparent to-gray-50 to-80%" />
+          <div className="pointer-events-none absolute inset-x-0 bottom-0 h-9 bg-linear-to-b from-transparent to-gray-50 to-80%" />
```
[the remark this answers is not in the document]
gabmontes: No. Is ok.

dvnahuel (approved):
LGTM!

ArturDolzan (dismissed):

web/src/components/walletDrawer/walletDrawerContent.tsx
```diff
@@ -139,14 +139,15 @@ export function WalletDrawerContent() {
           />
         </div>
         <div className="relative min-h-0 flex-1 overflow-hidden">
-          <div className="absolute inset-0 overflow-y-auto px-4 py-6 md:px-6">
+          <div className="absolute inset-0 overflow-y-auto px-4 pt-6 md:px-6">
             <ActivityList
+              className="pb-6"
```
gndelia: This is the only usage of `ActivityList`. So I think we can just move `pb-6` to its inside implementation

gndelia (changes requested):

web/src/components/walletDrawer/walletDrawerContent.tsx
gabmontes: That was the first implementation but I think it is better to keep the ActivityList component clean and manage all the paddings in the same place at the container.

gndelia: then I'd do `*:pb-6` in the parent container. I never like that of adding `className` props and leave concatenating random classes which may collide with the internal implementation.

gabmontes: The only problem with that is the empty state, ActivityListEmptyState. And that shows the current architecture is flawed. Both the container and the overlay are part of ActivityList, not part of the container. So I moved those `div` and it is much more clean now.

ArturDolzan (approved):

gndelia (approved):

PR #262, opened 2026-03-24
Add reserve buffer to yield allocation card

### LINKED ISSUE #238: Show reserve buffer

In the analytics page, the reserve buffer -amount not deployed to any strategy, used for instant withdrawal liquidity- shall be displayed.

### Proposed implementation

Add a last row to the allocations table in the right, to show the buffer. It should be equal to the difference between the TVL and the sum of all allocations (strategy debt amounts).

ArturDolzan:
### Description

This PR shows the idle treasury funds not deployed to any strategy as a last row in the yield allocation table. The buffer is computed as withdrawable minus totalDebt per token, converted to USD, and colored using the existing palette sequence.

### Screenshots

Captura de Tela 2026-03-24 às 16 14 45

### Related issue(s)

Closes #238 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/pages/analytics/utils.ts
+++ b/web/src/pages/analytics/utils.ts
@@ -15,7 +15,7 @@ const colorPalette = [
   "bg-pink-400",
 ];
 
-const assignColor = (index: number) =>
+export const assignColor = (index: number) =>
   colorPalette[index % colorPalette.length] ?? "bg-gray-400";
 
 const priceDecimals = 8;
@@ -86,23 +86,19 @@ export const toYieldItems = function ({
   return items;
 };
 
-// Returns the reserve buffer as an AllocationItem, or null if the buffer is zero or negative.
+// Returns the reserve buffer amount and label, or null if the buffer is zero or negative.
 // The buffer is the idle amount not deployed to any strategy, used for instant withdrawal liquidity.
-// Color is assigned after all strategy items, keeping the palette consistent.
-export const toReserveBufferItem = function ({
-  label,
+// Color is intentionally omitted — callers assign it based on their rendering context.
+export const toReserveBufferAmount = function ({
   treasuryTokens = [],
   whitelistedTokens = [],
 }: {
-  label: string;
   treasuryTokens?: TreasuryToken[];
   whitelistedTokens?: Token[];
 }) {
   let amount = 0;
-  let strategyCount = 0;
 
   for (const {
-    activeStrategies,
     latestPrice,
     tokenAddress,
     totalDebt,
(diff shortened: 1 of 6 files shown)


web/src/pages/analytics/utils.ts
```diff
@@ -85,3 +85,46 @@ export const toYieldItems = function ({
...
+    amount +=
+      (Number(formatUnits(BigInt(withdrawable), decimals)) -
+        Number(formatUnits(BigInt(totalDebt), decimals))) *
+      price;
+  }
+
+  if (amount <= 0) return null;
+
+  return { amount, color: assignColor(strategyCount), label };
+};
```
gndelia, line 117: It seems to me that this can be tested, particularly the amount calcs part. Thoughts?

ArturDolzan: Good point. Added on f7cde0585ef20506c1b07054b924e368aed97e42. Thanks!

gndelia (dismissed):

dvnahuel:
@ArturDolzan Let's make the height of both container the same. 
Screenshot 2026-03-24 at 3 42 42 PM

web/src/pages/analytics/utils.ts
```diff
@@ -85,3 +85,46 @@ export const toYieldItems = function ({
...
+    withdrawable,
+  } of treasuryTokens) {
+    const token = findToken(tokenAddress, whitelistedTokens);
+    const decimals = token?.decimals ?? 18;
+    const price = Number(formatUnits(BigInt(latestPrice), priceDecimals));
+
+    for (const { totalDebt: stratDebt } of activeStrategies) {
+      if (Number(formatUnits(BigInt(stratDebt), decimals)) * price > 0) {
+        strategyCount++;
+      }
```
gabmontes: Can we just test if `stratDebt` is nonzero? Regardless, this is only used to increment `strategyCount`. Perhaps a simpler approach would be to just calculate the amount and let the color be picked at rendering time in `Analytics`.

ArturDolzan: Applied on d888c1f23396e56a3ff5ac0f58fad4a32e243f06. Thanks!

ArturDolzan:
> @ArturDolzan Let's make the height of both container the same. Screenshot 2026-03-24 at 3 42 42 PM

@dvnahuel applied on d888c1f23396e56a3ff5ac0f58fad4a32e243f06. 

Captura de Tela 2026-03-24 às 21 17 19

gabmontes (dismissed):

web/src/pages/analytics/utils.ts
```diff
@@ -85,3 +85,34 @@ export const toYieldItems = function ({
...
+    const decimals = token?.decimals ?? 18;
+    const price = Number(formatUnits(BigInt(latestPrice), priceDecimals));
+
+    amount +=
+      (Number(formatUnits(BigInt(withdrawable), decimals)) -
+        Number(formatUnits(BigInt(totalDebt), decimals))) *
+      price;
+  }
+
+  return amount > 0 ? amount : null;
```
gndelia: I've got to ask - why not returning 0 instead of null?

ArturDolzan: Yes, sounds more natural I guess?

ArturDolzan: @gndelia suggestion applied on 0812bf77d970d068b93c997eecc876ffba029f04. Thanks!

gndelia (approved):

gabmontes (approved):

PR #305, opened 2026-04-07
api, web: update worker configs

joshuasing:
### Description

Update Cloudflare worker configs, add security headers to the API.

### Related issue(s)

Closes #
Related to #

### Checklist

- [x] Manual testing passed.


api/src/security-headers.ts
```diff
@@ -0,0 +1,65 @@
...
+  c.header("Cross-Origin-Opener-Policy", "same-origin");
+  c.header("Cross-Origin-Resource-Policy", "same-origin");
+  c.header("Origin-Agent-Cluster", "?1");
+  c.header("Permissions-Policy", permissionsPolicy);
+  c.header("Referrer-Policy", "no-referrer");
+  c.header("X-Content-Type-Options", "nosniff");
+  c.header("X-Frame-Options", "DENY");
+
+  if (!c.res.headers.has("Cache-Control")) {
+    c.header("Cache-Control", "no-store");
```
gndelia: @gabmontes could this conflict with the Wrangler cache we're using?

api/src/index.ts
```diff
@@ -7,6 +7,7 @@ import * as borrow from "./borrow.ts";
 import { convertBigIntsToString } from "./convert-bigints-to-string.ts";
 import { getSubgraphUrl } from "./env.ts";
 import { validateAddress, validateParam } from "./param-validators.ts";
+import { securityHeaders } from "./security-headers.ts";
```
gabmontes, line 10: What about using https://hono.dev/docs/middleware/builtin/secure-headers?

api/src/security-headers.ts
gabmontes: Actually, this is applied after the route handlers run and only if `Cache-Control` was not set. So it seems safe.

api/src/security-headers.ts
```diff
@@ -0,0 +1,65 @@
...
+  "xr-spatial-tracking=()",
+].join(", ");
+
+export const securityHeaders: MiddlewareHandler = async function (c, next) {
+  await next();
+
+  c.header("Content-Security-Policy", contentSecurityPolicy);
+  c.header("Cross-Origin-Embedder-Policy", "require-corp");
+  c.header("Cross-Origin-Opener-Policy", "same-origin");
+  c.header("Cross-Origin-Resource-Policy", "same-origin");
```
gndelia: I think this will block requests from the web, as they will live in a different subdomains

joshuasing: I believe it will correctly allow CORS `fetch`es from `app.vetro.org` - however `` or non-CORS fetches would be blocked.

This could be changed to `same-site`, which I believe would allow from `*.vetro.org`.

api/src/index.ts
joshuasing: Thanks, good catch. I found that and then apparently forgot about it 😅

api/src/security-headers.ts
joshuasing: I will remove this for now, I don't believe it has a lot of benefit as no authenticated requests are handled.

api/src/security-headers.ts
gndelia: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Same-origin_policy#definition_of_an_origin) `http://store.company.com/dir/page.html` and `http://news.company.com/dir/page.html` would fail on "same-origin" because the whole host differs (they are on different subdomains), which is similar to our case

joshuasing: I am a little confused by how this is actually supposed to work, haha.

MDN for CORP header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Resource-Policy) states:
> The HTTP Cross-Origin-Resource-Policy [response header](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) (CORP) indicates that the browser should block [no-cors](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#no-cors) cross-origin or cross-site requests to the given resource.

As the API handles CORS, and no-cors requests shouldn't be made from any other origin, I _believe_ `same-origin` will work and would be safer than `same-site` (which would allow no-cors requests)

api/src/security-headers.ts
```diff
@@ -0,0 +1,71 @@
...
+  "pictureInPicture",
+  "publickeyCredentialsGet",
+  "screenWakeLock",
+  "serial",
+  "storageAccess",
+  "syncXhr",
+  "usb",
+  "webShare",
+  "windowManagement",
+  "xrSpatialTracking",
```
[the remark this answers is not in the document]
joshuasing: These values are taken from the `StandardizedFeatures` type in hono:
```typescript
type StandardizedFeatures = 'accelerometer' | 'ambientLightSensor' | 'attributionReporting' | 'autoplay' | 'battery' | 'bluetooth' | 'camera' | 'chUa' | 'chUaArch' | 'chUaBitness' | 'chUaFullVersion' | 'chUaFullVersionList' | 'chUaMobile' | 'chUaModel' | 'chUaPlatform' | 'chUaPlatformVersion' | 'chUaWow64' | 'computePressure' | 'crossOriginIsolated' | 'directSockets' | 'displayCapture' | 'encryptedMedia' | 'executionWhileNotRendered' | 'executionWhileOutOfViewport' | 'fullscreen' | 'geolocation' | 'gyroscope' | 'hid' | 'identityCredentialsGet' | 'idleDetection' | 'keyboardMap' | 'magnetometer' | 'microphone' | 'midi' | 'navigationOverride' | 'payment' | 'pictureInPicture' | 'publickeyCredentialsGet' | 'screenWakeLock' | 'serial' | 'storageAccess' | 'syncXhr' | 'usb' | 'webShare' | 'windowManagement' | 'xrSpatialTracking';
```

api/src/security-headers.ts
gndelia: copilot seems to believe the same https://github.com/vetro-protocol/vetro-monorepo/pull/305#discussion_r3047586279

Regardless, if you believe it will work, I'm ok with pushing this
we can always switch it if we get rejected CORS requests from the app to the api

gabmontes (approved):
LGTM

gndelia (approved):

PR #310, opened 2026-04-07
Fix API url for landing

gndelia:
### Description

Locally, with wrangler, site and API are served in the same url. But we deployed the API and WEB in different subdomains, so using a relative path does not work. This PR should fix that

Note that's untested because I cannot hit the API from localhost

### Checklist

- [x] Manual testing passed.


gndelia:
Incorrectly created - I misunderstood the implementation

Issue #239, opened 2026-03-19
Support vetBTC and svetBTC


gabmontes:
### Feature or change description

There will be another set of tokens pegged to BTC instead of USD: vetBTC and the staked version, svetBTC. All the functionalities in the app should be modified to integrate these tokens too.

### Proposed implementation

We can take some inspiration from these pages:

https://app.avantprotocol.com/transparency
https://www.circle.com/transparency

gndelia:
Taking some notes (so I don't forget)

- [ ] The `/variable-stake/apy` needs to be extended for several earn pools - 
- [x] We'll need (Eventually) deploy the subgraph with the new Earn pool https://github.com/vetro-protocol/vetro-monorepo/pull/357
- [ ]  The `/variable-stake/rewards/${address}` endpoint will need to return rewards for all Staking Vaults (or accept the vault address) See https://github.com/vetro-protocol/vetro-monorepo/issues/359
- [x] Add the "From Pools" hover section in the Earn stats section https://github.com/vetro-protocol/vetro-monorepo/pull/349
- [x] Update "Activities" in the Activity drawer to reflect the token / vaults operated (if needed)
- [x] Define how we solve the "Claim all" in the exit queue for Earn https://github.com/vetro-protocol/vetro-monorepo/issues/355
-  [x] The "Send to Queue" texts should list only whitelisted tokens associated to the selected vetro token

<img width="616" height="605" alt="Image" src="https://github.com/user-attachments/assets/029aeb32-6af7-42b1-8641-791d1aad509c" />

- [x] The swap form, when sending `vetro` tokens to the redeem queue, lists all whitelisted tokens in existence for all vetro tokens

<img width="458" height="413" alt="Image" src="https://github.com/user-attachments/assets/39ffd959-99e5-4fa9-a12b-2afe00d8771e" />

gndelia:
contracts added https://github.com/vetro-protocol/vetro-contracts/issues/111#issuecomment-4295485845

PR #354, opened 2026-04-23
Index sVetBTC staking vault in subgraph

gndelia:
### Description

Add a second data source to the subgraph so it indexes the newly deployed sVetBTC staking vault at `0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e` (starts at block `24931128`) alongside the existing sVUSD vault. The existing `StakingVault` data source is renamed to `sVusdStakingVault` so both vaults have asset-specific, unambiguous names in `networks.json`, `subgraph.yaml`, and the `generated/` tree. Both vaults share the same ABI and the existing mapping file (`src/earn.ts`) — every handler already routes per-vault via `dataSource.address()`, so no code changes were needed there beyond the import path update. Subgraph version bumped to `1.3.0`.

### Screenshots

N/A — subgraph-only change.

### Related issue(s)

Related to #239

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/subgraph/networks.json
+++ b/subgraph/networks.json
@@ -1,8 +1,12 @@
 {
   "mainnet": {
-    "StakingVault": {
+    "sVusdStakingVault": {
       "address": "0x476310E34D2810f7d79C43A74E4D79405bd7a925",
       "startBlock": 24625396
+    },
+    "sVetBTCStakingVault": {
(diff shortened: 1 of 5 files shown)


gabmontes (commented):
Let's close it in favor of #357.

PR #376, opened 2026-04-30
Return APY per staking vault

### LINKED ISSUE #360: Calculate APY per staking pool

The current `/variable-stake/apy` endpoint only works for 1 staking vault, but it should be filtered per staking vault so each one returns its own APY (or return an object with APYs for all staking vaults).

`useApy` is already used in `PoolInfoBar` which only renders one pool

Depends on #357 being merged first

Related to #239

gndelia:
### Description

`/variable-stake/apy` previously aggregated `VaultHistory` records across all staking vaults and returned a single APY. With multiple vaults live (sVUSD, sVETBTC) and `PoolInfoBar` rendering one pool at a time, callers need APY per pool.

The endpoint now returns `Record<Address, { "7d": number }>` keyed by vault address, by filtering `vaultHistories` with `stakingVaultAddress_in` and grouping per vault. Every configured vault is always present in the response, with `{ "7d": 0 }` when there is not enough history to compute a slope. `useApy(stakingVaultAddress)` keeps a single shared cache and uses React Query's `select` to extract its slice — mirroring the pattern already used by `useUserRewards`.

### Screenshots

image

(totally made up numbers)

### Related issue(s)

Closes #360

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -102,7 +102,7 @@ Returns the APY for each supported staking vault, indexed by staking vault addre
 ```jsonc
 {
   "0x476310E34D2810f7d79C43A74E4D79405bd7a925": {
-    "7d": 100, // 100 = 1%
+    "7d": 4.21,
   },
   "0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e": {
     "7d": 0,
--- a/api/README.md
+++ b/api/README.md
@@ -95,13 +95,18 @@ Gets the amount of collateral assets in a given Morpho market.
 
 ### `GET /variable-stake/apy`
 
-Returns the APY of the Vetro vault calculated using the share value variations over the last 7 days.
+Returns the APY for each supported staking vault, indexed by staking vault address, calculated using the share value variations over the last 7 days. Each supported staking vault has an entry in the response; vaults with insufficient history return `{ "7d": 0 }`.
 
 #### Sample Response
 
 ```jsonc
 {
-  "7d": 100, // 100 = 1%
+  "0x476310E34D2810f7d79C43A74E4D79405bd7a925": {
+    "7d": 100, // 100 = 1%
+  },
+  "0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e": {
+    "7d": 0,
+  },
 }
 ```
 
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -168,7 +168,9 @@ app.get(
   async function (c) {
     try {
       const url = getSubgraphUrl(c.env);
-      const data = await variableStake.getApy({ url });
+      const data = await variableStake.getApy({
+        url,
+      });
       return c.json(data);
     } catch (error) {
       throw new Error(`Failed to get APY: ${error.message}`);
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -168,9 +168,7 @@ app.get(
   async function (c) {
     try {
       const url = getSubgraphUrl(c.env);
-      const data = await variableStake.getApy({
-        url,
-      });
+      const data = await variableStake.getApy({ url });
       return c.json(data);
     } catch (error) {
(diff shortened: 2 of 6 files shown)


api/test/variable-stake.test.ts
```diff
@@ -21,6 +22,121 @@ vi.mock("../src/merkl.ts", () => ({
   getUserRewards: vi.fn(),
 }));
 
+vi.mock("../src/graphql.ts", () => ({
+  runQuery: vi.fn(),
+}));
+
```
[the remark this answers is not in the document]
gndelia: Fixed in [da4a739](https://github.com/vetro-protocol/vetro-monorepo/pull/376/commits/da4a7399c72f211807fb582f9bdd156053728185)

api/README.md
```diff
@@ -95,13 +95,18 @@ Gets the amount of collateral assets in a given Morpho market.
...
-Returns the APY of the Vetro vault calculated using the share value variations over the last 7 days.
+Returns the APY for each supported staking vault, indexed by staking vault address, calculated using the share value variations over the last 7 days. Each supported staking vault has an entry in the response; vaults with insufficient history return `{ "7d": 0 }`.
 
 #### Sample Response
 
 ```jsonc
 {
-  "7d": 100, // 100 = 1%
+  "0x476310E34D2810f7d79C43A74E4D79405bd7a925": {
+    "7d": 100, // 100 = 1%
```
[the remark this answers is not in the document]
gndelia: Fixed in [436589c](https://github.com/vetro-protocol/vetro-monorepo/pull/376/commits/436589c7ae2607d57890e71e72778694fdd3d5dc). The value is the percentage directly (e.g. 4.21 = 4.21%); updated the sample to reflect that.

ArturDolzan (dismissed):

web/src/hooks/useApy.ts
```diff
@@ -1,21 +1,19 @@
...
     enabled: apiUrl !== undefined && isValidUrl(apiUrl),
     queryFn: () =>
-      fetch(`${apiUrl}/variable-stake/apy`).then(
-        (data: ApyResponse) => data["7d"],
-      ),
+      fetch(`${apiUrl}/variable-stake/apy`) as Promise<ApyResponse>,
     queryKey: ["variable-stake-apy"],
     refetchInterval: 5 * 60 * 1000, // 5 minutes
     retry: 2,
+    select: (data) => data[stakingVaultAddress]?.["7d"] ?? 0,
```
gabmontes: If the API response always contain the vault and the "7d" property, it can be safely obtained without need for guards and defaults.

api/src/index.ts
```diff
@@ -168,7 +168,9 @@ app.get(
   async function (c) {
     try {
       const url = getSubgraphUrl(c.env);
-      const data = await variableStake.getApy({ url });
+      const data = await variableStake.getApy({
+        url,
+      });
```
gabmontes: It looks uglier now...

gndelia: Beauty is relative (?)

I guesss this happened because as part of the implementation I added a 2nd parameter (taking this call into a 2-liner), and then I removed that extra param, but prettier preferred to keep it like this... I'll fix it

web/src/hooks/useApy.ts
gndelia: Fixed in [3423e1e](https://github.com/vetro-protocol/vetro-monorepo/pull/376/commits/3423e1e4de8eff3678ae41d7bf4a0ff58c489748).

api/src/index.ts
gndelia: Fixed in [f9eea58](https://github.com/vetro-protocol/vetro-monorepo/pull/376/commits/f9eea583661189155f6a24b188208e5a082c1f6e).

gabmontes (approved):

ArturDolzan (approved):

PR #405, opened 2026-05-08
Add exit ticket ready-to-withdraw notification

### LINKED ISSUE #395: Add notifications for ready to exit tickets on Earn

The app currently shows notifications for 

- Pegged Tokens ready to redeem (After the 6/7 block delay)
- Positions from `/borrow` that are not healthy enough and are in danger of liquidation

There's another process that's also "delayed", and it's the Exit queue of Earn page. In this case, it takes 7 days or so (configurable) to exit. After that, users need to claim each one (or claim all at once - see #355 )

We need to add a notification per exit ticket that's "Ready" to be claimed.

The UI should, upon clicking, take the user to the Earn page, scrolling to the Exit table (just as we do for the borrow and swap page)

Text is TBD

gndelia:
### Description

Surface a global notification when at least one exit ticket has finished its cooldown and is ready to claim, mirroring the existing "ready to redeem" notification pattern. Clicking the notification scrolls to the exit-tickets section on the Earn page, or navigates there if the user is on a different page.

Along the way:

- Extracts the timestamp-flip-tick logic from `useRedeemItems` into a shared `useNowTickingPast` hook so both the redeem and exit-ticket notifications can flip from "pending" to "ready" the moment a cooldown matures, without polling.
- Adds an `id="exit-tickets"` anchor on the section so the notification can scroll to it.
- Replaces an inlined `Math.floor(Date.now() / 1000)` in `getTicketStatus` with the existing `unixNowTimestamp` helper, so the now-source is consistent across the codebase.
- Adds a small `useI18nNavigate` helper that the new notification uses to navigate while preserving the current locale.

**Commits:**

- 20e5c08 Add useI18nNavigate hook
- 0880923 Add exit ticket ready-to-withdraw notification
- b89ad2a Use unixNowTimestamp in getTicketStatus

### Screenshots

image

a notification like this now appears

### Related issue(s)

Closes #395

### Checklist

- [ ] Manual testing passed.


gabmontes (dismissed):
LGTM

ArturDolzan (dismissed):

ArturDolzan (approved):

gabmontes (approved):

PR #410, opened 2026-05-12
Add worker entry point to enforce CSP

gabmontes:
### Description

Add a new worker entry point in src/index.ts to handle security headers and asset fetching.

This pull request updates the Cloudflare Workers type definitions to the latest version and introduces a new Cloudflare Worker entry point for the web application that proxies static assets and injects enhanced security headers. The changes primarily affect dependency management and the security of served HTML content.

**Dependency updates:**

* Upgraded `@cloudflare/workers-types` from version `4.20260212.0` to `4.20260504.1` throughout the project, including in `pnpm-lock.yaml`, `web/package.json`, and transitive dependencies for `wrangler` and `@cloudflare/vite-plugin`. This ensures compatibility with the latest Cloudflare Workers APIs. [[1]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL97-R97) [[2]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL115-R115) [[3]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL358-R361) [[4]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL412-R415) [[5]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL652-R656) [[6]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL8871-R8880) [[7]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL8929-R8932) [[8]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL18203-R18204) [[9]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL18214-R18221) [[10]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL18231-R18238) [[11]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR18249) [[12]](diffhunk://#diff-b861012a5dd72b8a9f3281b7cf09f5a779c98569d040b1bbc1db50f1b15e7cceR47)

**Web worker improvements:**

* Added a new worker entry point in `web/src/index.ts` that proxies static assets and injects strict security headers (including a dynamically generated Content Security Policy) on HTML responses to improve security and prevent common web vulnerabilities.
* Updated `web/wrangler.jsonc` to set `src/index.ts` as the main entry point for the worker and bind the static asset namespace as `ASSETS`, enabling the new worker logic. [[1]](diffhunk://#diff-189cd30f7e4db147718767a051d2fbdd87db84109122913509a1e7299d08835aR3) [[2]](diffhunk://#diff-189cd30f7e4db147718767a051d2fbdd87db84109122913509a1e7299d08835aR18)

**Other dependency adjustments:**

* Updated `eslint-module-utils` to use a newer version of `@typescript-eslint/parser` in the lockfile, ensuring compatibility with other linting tools. [[1]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL14167-L14174) [[2]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL14195-R14196)

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/index.ts
+++ b/web/src/index.ts
@@ -1,0 +1,134 @@
+// Worker entry point – proxies static assets and injects security headers.
+
+/// <reference types="@cloudflare/workers-types" />
+
+import { allChains } from "networks";
+import { getUrlOrigin } from "utils/url";
+
+type Env = {
+  ASSETS: Fetcher;
+};
+
+// Extract unique RPC URLs
+const rpcUrls = allChains.flatMap((chain) => chain.rpcUrls.default.http);
+
+const allUrls: (string | undefined)[] = [
+  import.meta.env.VITE_PORTAL_API_URL,
+  import.meta.env.VITE_VETRO_API_URL,
+  ...rpcUrls,
+  import.meta.env.VITE_SENTRY_DSN,
+];
+
+// Build connect-src dynamically from env vars baked in at build time.
+const connectSrc = [
+  "'self'",
+  ...allUrls.map(getUrlOrigin).filter(Boolean),
+].join(" ");
+
+const workerSrc = import.meta.env.VITE_SENTRY_DSN ? "blob:" : "'none'";
+
+// Deny all permissions – mirrors api/src/security-headers.ts.
+const permissionsPolicy = [
+  "accelerometer",
+  "ambient-light-sensor",
+  "attribution-reporting",
+  "autoplay",
+  "battery",
+  "bluetooth",
+  "camera",
+  "ch-ua",
+  "ch-ua-arch",
+  "ch-ua-bitness",
+  "ch-ua-full-version",
+  "ch-ua-full-version-list",
+  "ch-ua-mobile",
+  "ch-ua-model",
+  "ch-ua-platform",
+  "ch-ua-platform-version",
+  "ch-ua-wow64",
+  "compute-pressure",
+  "cross-origin-isolated",
+  "direct-sockets",
+  "display-capture",
+  "encrypted-media",
+  "execution-while-not-rendered",
+  "execution-while-out-of-viewport",
+  "fullscreen",
+  "geolocation",
+  "gyroscope",
+  "hid",
+  "identity-credentials-get",
+  "idle-detection",
+  "keyboard-map",
+  "magnetometer",
+  "microphone",
+  "midi",
+  "navigation-override",
+  "payment",
+  "picture-in-picture",
+  "publickey-credentials-get",
+  "screen-wake-lock",
+  "serial",
+  "storage-access",
+  "sync-xhr",
+  "usb",
+  "web-share",
+  "window-management",
+  "xr-spatial-tracking",
+]
+  .map((feature) => `${feature}=()`)
+  .join(", ");
+
+const csp = [
+  "base-uri 'none'",
+  `connect-src ${connectSrc}`,
+  "default-src 'none'",
+  "font-src 'self' https://fonts.gstatic.com",
+  "form-action 'none'",
+  "frame-ancestors 'none'",
+  "img-src 'self' data: https://hemilabs.github.io",
+  "script-src 'self'",
+  // Tailwind v4 injects styles via a <style> tag, so 'unsafe-inline' is needed.
+  "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
+  "upgrade-insecure-requests",
+  `worker-src ${workerSrc}`,
+].join("; ");
+
+// Applied to all responses – these have meaningful effect on sub-resources.
+const commonHeaders = {
+  "Content-Security-Policy": "default-src 'none'",
+  "Cross-Origin-Resource-Policy": "same-origin",
+  "Referrer-Policy": "strict-origin-when-cross-origin",
+  "X-Content-Type-Options": "nosniff",
+};
+
+// Applied only to HTML documents – these are navigation-level controls.
+// CSP is stricter for HTML responses so it overrides the commonHeaders CSP.
+const htmlHeaders = {
+  "Content-Security-Policy": csp,
+  "Cross-Origin-Opener-Policy": "same-origin",
+  "Origin-Agent-Cluster": "?1",
+  "Permissions-Policy": permissionsPolicy,
+  "X-DNS-Prefetch-Control": "off",
+  "X-Frame-Options": "DENY",
+};
+
+export default {
+  async fetch(request: Request, env: Env) {
+    const response = await env.ASSETS.fetch(request);
+    const newResponse = new Response(response.body, response);
+
+    for (const [key, value] of Object.entries(commonHeaders)) {
+      newResponse.headers.set(key, value);
+    }
+
+    const contentType = response.headers.get("Content-Type") ?? "";
+    if (contentType.includes("text/html")) {
+      for (const [key, value] of Object.entries(htmlHeaders)) {
+        newResponse.headers.set(key, value);
+      }
+    }
+
+    return newResponse;
+  },
+} satisfies ExportedHandler<Env>;
--- a/web/wrangler.jsonc
+++ b/web/wrangler.jsonc
@@ -1,5 +1,6 @@
 {
   "$schema": "./node_modules/wrangler/config-schema.json",
+  "main": "src/index.ts",
   "name": "vetro-web-staging",
   "compatibility_date": "2026-02-06",
   "workers_dev": false,
@@ -14,6 +15,7 @@
     },
   },
   "assets": {
+    "binding": "ASSETS",
     "not_found_handling": "single-page-application",
   },
   "env": {
--- a/web/wrangler.jsonc
+++ b/web/wrangler.jsonc
@@ -2,7 +2,7 @@
   "$schema": "./node_modules/wrangler/config-schema.json",
   "main": "src/index.ts",
   "name": "vetro-web-staging",
-  "compatibility_date": "2026-02-06",
+  "compatibility_date": "2026-02-05",
   "workers_dev": false,
   "preview_urls": true,
   "upload_source_maps": true,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -94,7 +94,7 @@ importers:
         version: 4.0.18(@opentelemetry/api@1.9.0)(@types/node@24.10.13)(jiti@2.6.1)(lightningcss@1.30.2)(yaml@2.8.2)
       wrangler:
         specifier: 4.65.0
-        version: 4.65.0(@cloudflare/workers-types@4.20260212.0)(bufferutil@4.1.0)(utf-8-validate@5.0.10)
+        version: 4.65.0(@cloudflare/workers-types@4.20260504.1)(bufferutil@4.1.0)(utf-8-validate@5.0.10)
 
   landing:
     devDependencies:
@@ -112,7 +112,7 @@ importers:
         version: 5.9.3
       wrangler:
         specifier: 4.79.0
-        version: 4.79.0(bufferutil@4.1.0)(utf-8-validate@5.0.10)
+        version: 4.79.0(@cloudflare/workers-types@4.20260504.1)(bufferutil@4.1.0)(utf-8-validate@5.0.10)
 
   packages/bridge:
     dependencies:
@@ -355,7 +355,10 @@ importers:
     devDependencies:
       '@cloudflare/vite-plugin':
         specifier: 1.23.1
-        version: 1.23.1(bufferutil@4.1.0)(utf-8-validate@5.0.10)(vite@7.3.1(@types/node@24.1.0)(jiti@2.6.1)(lightningcss@1.30.2)(yaml@2.8.2))(workerd@1.20260205.0)(wrangler@4.63.0(@cloudflare/workers-types@4.20260212.0)(bufferutil@4.1.0)(utf-8-validate@5.0.10))
+        version: 1.23.1(bufferutil@4.1.0)(utf-8-validate@5.0.10)(vite@7.3.1(@types/node@24.1.0)(jiti@2.6.1)(lightningcss@1.30.2)(yaml@2.8.2))(workerd@1.20260205.0)(wrangler@4.63.0(@cloudflare/workers-types@4.20260504.1)(bufferutil@4.1.0)(utf-8-validate@5.0.10))
+      '@cloudflare/workers-types':
+        specifier: 4.20260504.1
+        version: 4.20260504.1
       '@sentry/vite-plugin':
         specifier: 5.2.0
         version: 5.2.0(encoding@0.1.13)(rollup@4.55.3)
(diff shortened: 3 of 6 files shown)


web/wrangler.jsonc
```diff
@@ -14,6 +15,7 @@
     },
   },
   "assets": {
+    "binding": "ASSETS",
```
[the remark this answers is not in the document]
gabmontes: No. This project uses @cloudflare/vite-plugin. The plugin owns the Vite build pipeline, outputs assets into its own managed location, and injects the correct assets.directory into Wrangler's configuration at build time automatically.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// Build connect-src dynamically from env vars baked in at build time.
+const connectSrc = [
+  "'self'",
+  new URL(import.meta.env.VITE_PORTAL_API_URL).origin,
+  new URL(import.meta.env.VITE_VETRO_API_URL).origin,
+  ...rpcOrigins.map((url) => new URL(url).origin),
+  import.meta.env.VITE_SENTRY_DSN
+    ? new URL(import.meta.env.VITE_SENTRY_DSN).origin
+    : "",
```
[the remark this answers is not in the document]
gabmontes: No. This project is built with Vite. This is ok.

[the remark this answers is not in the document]
gabmontes: Ok. Improved.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
+// Worker entry point – proxies static assets and injects security headers.
+
+/// <reference types="@cloudflare/workers-types" />
+
+import { mainnet } from "viem/chains";
+
+interface Env {
+  ASSETS: Fetcher;
+}
```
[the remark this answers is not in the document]
gabmontes: Ok. Changed.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (!contentType.includes("text/html")) {
+      return response;
+    }
+
+    const newResponse = new Response(response.body, response);
+    for (const [key, value] of Object.entries(securityHeaders)) {
+      newResponse.headers.set(key, value);
+    }
+    return newResponse;
+  },
```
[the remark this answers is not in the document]
gabmontes: Don't worry. It works.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  async fetch(request: Request, env: Env) {
+    const response = await env.ASSETS.fetch(request);
+
+    // Only inject security headers on HTML document responses. Browsers enforce
+    // CSP on navigation only, not on sub-resources. Non-HTML responses (JS,
+    // CSS, fonts, images) are returned unmodified.
+    const contentType = response.headers.get("Content-Type") ?? "";
+    if (!contentType.includes("text/html")) {
+      return response;
+    }
```
joshuasing: This behaviour is correct for the `Content-Security-Policy` header, but all other headers should be set for non-HTML responses.

I think it would be okay to set `Content-Security-Policy` on all responses as well, since it will be ignored by the browser when non-applicable. This would also mean if it can be applied to a page, it will be, versus it being gated here.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+const csp = [
+  "base-uri 'self'",
+  `connect-src ${connectSrc}`,
+  "default-src 'none'",
+  "font-src 'self' https://fonts.gstatic.com",
+  "form-action 'none'",
+  "frame-ancestors 'none'",
+  "img-src 'self' data: https://hemilabs.github.io",
+  "script-src 'self'",
+  "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
```
joshuasing, line 92: Is `unsafe-inline` required here? If so, could we add a comment explaining what causes it to be needed?

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  "upgrade-insecure-requests",
+  `worker-src ${workerSrc}`,
+].join("; ");
+
+const securityHeaders = {
+  "Content-Security-Policy": csp,
+  "Cross-Origin-Opener-Policy": "same-origin",
+  "Cross-Origin-Resource-Policy": "same-origin",
```
joshuasing: `Cross-Origin-Embedder-Policy` ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy)) should be set as well to restrict loading and embedding cross-origin resources.

```suggestion
  "Cross-Origin-Embedder-Policy": "credentialless",
  "Cross-Origin-Opener-Policy": "same-origin",
  "Cross-Origin-Resource-Policy": "same-origin",
```

Using a value of `require-corp` would be more preferable, however I believe WalletConnect iframes would break.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  "Referrer-Policy": "strict-origin-when-cross-origin",
+  "X-Content-Type-Options": "nosniff",
```
joshuasing, line 102: I think we should add [`Origin-Agent-Cluster`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin-Agent-Cluster) and [`X-DNS-Prefetch-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control) here to match the API security headers. Also note, `X-Content-Type-Options` should already be set by Cloudflare, however it does not hurt to set here as well.
```suggestion
  "Origin-Agent-Cluster": "?1",
  "X-Content-Type-Options": "nosniff",
  "X-DNS-Prefetch-Control": "off",
```

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  "form-action 'none'",
+  "frame-ancestors 'none'",
+  "img-src 'self' data: https://hemilabs.github.io",
+  "script-src 'self'",
+  "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
+  "upgrade-insecure-requests",
+  `worker-src ${workerSrc}`,
+].join("; ");
+
+const securityHeaders = {
```
joshuasing: In a future pull request, it would be ideal to add `Permissions-Policy` to block all permissions requests - similar to https://github.com/vetro-protocol/vetro-monorepo/blob/8c49efd2f28482ec6a40ebb23dc80d50647a4279/api/src/security-headers.ts

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+// Build connect-src dynamically from env vars baked in at build time.
+const connectSrc = [
+  "'self'",
+  ...allUrls.map((url) => new URL(url).origin),
+].join(" ");
+
+const workerSrc = import.meta.env.VITE_SENTRY_DSN ? "blob:" : "'none'";
+
+const csp = [
+  "base-uri 'self'",
```
joshuasing: I believe `base-url` can be set to `none` here, which prevents an injected `<base>` tag from changing the relative URL on the page (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/base-uri):
```suggestion
  "base-uri 'none'",
```

joshuasing (changes requested):
Thank you! Some comments/suggestions below - additional headers could be done in a separate PR if wanted.

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type Env = {
+  ASSETS: Fetcher;
+};
+
+// Extract unique URLs from "+"-separated RPC URLs, falling back to viem's
+// default.
+const rpcUrls: string[] = import.meta.env.VITE_RPC_URL_MAINNET
+  ? import.meta.env.VITE_RPC_URL_MAINNET.split("+")
+  : mainnet.rpcUrls.default.http;
```
gndelia: I recommend here using [allChains](https://github.com/vetro-protocol/vetro-monorepo/blob/02454a5ea69734992c8edccbcc61c7f22f8413a9/web/src/networks/index.ts#L6) because

1. The bridge feature requires querying RPCs from chains other than mainnet (like Hemi, Arbitrum, etc)
2. `allChains` includes the `mainnet` version that's already overwritten from a custom RPC url - so you can use `chain.rpcUrls.default.http`  on all of them

```suggestion
// Extract unique RPC URLs
const rpcUrls = allChains.flatMap(chain => chain.rpcUrls.default.http);
```

gndelia (changes requested):

web/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const allUrls = [
+  import.meta.env.VITE_PORTAL_API_URL,
+  import.meta.env.VITE_VETRO_API_URL,
+  ...rpcUrls,
+  import.meta.env.VITE_SENTRY_DSN,
+].filter(Boolean);
```
gndelia: If we use [isValidUrl](https://github.com/vetro-protocol/vetro-monorepo/blob/02454a5ea69734992c8edccbcc61c7f22f8413a9/web/src/utils/url.ts#L3) here, we cover ourselves from wrong configurations here

```suggestion
].filter(isValidUrl);
```

it all depends on the type of error you prefer to return

web/src/index.ts
gabmontes: Done. X-Content-Type-Options and Cross-Origin-Resource-Policy now apply to all responses. CSP, COOP, Referrer-Policy, and X-Frame-Options remain HTML-only as they should.

web/src/index.ts
gabmontes: It is required for Tailwind to work. Added a comment explaining so.

web/src/index.ts
gabmontes: COEP's primary value is enabling cross-origin isolation to unlock "SharedArrayBuffer" and precision timers and we don't use those. The existing "CSP", "CORP: same-origin" and "COOP: same-origin" already provide strong isolation.

And yes, "require-corp" would break WalletConnect's QR modal.

web/src/index.ts
gabmontes: Added.

web/src/index.ts
gabmontes: Ok. Added.

web/src/index.ts
gabmontes: Done.

web/src/index.ts
gabmontes: Done!

web/src/index.ts
gabmontes: Improved!

web/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+// Extract unique RPC URLs
+const rpcUrls = allChains.flatMap((chain) => chain.rpcUrls.default.http);
+
+const allUrls: (string | undefined)[] = [
+  import.meta.env.VITE_PORTAL_API_URL,
+  import.meta.env.VITE_VETRO_API_URL,
+  ...rpcUrls,
+  import.meta.env.VITE_SENTRY_DSN,
+];
```
gndelia, line 20: I'm fairly sure that WalletConnect (through rainbow kit) hits other endpoints as well. Yoy may want to compare with the portal config for this

gndelia: however, we never set the wallet connect proejct id, so this may be ok. But if we want to leave this configured for the future, better do it now

gabmontes: I say we address this the day we enable WalletConnect.

gndelia (dismissed):

web/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+export default {
+  async fetch(request: Request, env: Env) {
+    const response = await env.ASSETS.fetch(request);
+    const newResponse = new Response(response.body, response);
+
+    for (const [key, value] of Object.entries(commonHeaders)) {
+      newResponse.headers.set(key, value);
+    }
+
+    const contentType = response.headers.get("Content-Type") ?? "";
```
joshuasing, line 125: I think we need to set all of the headers for all responses.

It is far better for the browser to ignore non-applicable headers than for us to not set them entirely.

gabmontes: @joshuasing These seven headers are document/navigation-level controls. Browsers literally ignore them on non-HTML responses. There is no point in adding them to every other response.

joshuasing: That blanket characterisation isn't accurate. Of the seven headers in `htmlHeaders`, at least two are recommended for use on non-HTML responses:
- `Referrer-Policy` - [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#referrer-policy
) states: "we suggest forcing this behavior by sending this header on all responses."
- `Content-Security-Policy` - MDN explicitly states: "It should be set on all responses to all requests, not just the main document." (https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP). [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#1-content-security-policy-header) also states: "Send it in all HTTP responses, not just the index page."

For the remaining five - there is no downside to including them. If a browser chooses to act on a header, that should be its decision, not ours. If a response is not detected as HTML here, but exploits something one of these headers would prevent, would it not be best to have the header anyway?

gabmontes: @joshuasing You are right. Moved Referrer-Policy and applied CSP to non-HTML resources with a simpler `default-src 'none'` that would suffice and is more correct than the full CSP (which would not be wrong but seems wasteful).

I still don't think the other five are needed. `X-Content-Type-Options: nosniff` is already in `commonHeaders` and is specifically designed to prevent MIME confusion, content-type sniffing, etc.

joshuasing: Sounds good. For what it's worth, I'd still lean toward applying the other five uniformly as a defence-in-depth habit, but happy with this solution. Thanks for working through it.

joshuasing (dismissed):

gndelia:
@gabmontes conflicts!

gabmontes:
Conflicts solved.

gndelia (approved):

joshuasing (approved):

PR #444, opened 2026-06-03
Store total assets per day in vault history

gndelia:
### Description

This extends the subgraph to track historic TVL for each staking vault.

The block handler already snapshots `shareValue` once per UTC day on each `VaultHistory` row. This extends that same row with `totalAssets`, so a vault's total deposits — `totalAssets()`, denominated in the underlying pegged token — can be charted over time. No new entity or extra rows: blocks within a day keep overwriting the day's row, so only the last value per UTC day persists.

Because the block handler runs on every block during sync, redeploying backfills `totalAssets` for every past day automatically — no separate migration.

Changes:
- Add `totalAssets()` to the `StakingVault` ABI.
- Add a `totalAssets` field to the `VaultHistory` entity.
- Write `entity.totalAssets = vault.totalAssets()` in the block handler.
- Extend the `handleBlock` tests to cover it.
- Bump the subgraph version to 1.4.0.

### Screenshots

N/A — no UI changes.

### Related issue(s)

Related to https://github.com/hemilabs/ui-monorepo/issues/1967

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraph/abis/StakingVault.json
+++ b/subgraph/abis/StakingVault.json
@@ -51,6 +51,19 @@
     "stateMutability": "view"
   },
   {
+    "type": "function",
+    "name": "totalAssets",
+    "inputs": [],
+    "outputs": [
+      {
+        "name": "",
+        "type": "uint256",
+        "internalType": "uint256"
+      }
+    ],
+    "stateMutability": "view"
+  },
+  {
     "type": "event",
     "name": "Deposit",
     "inputs": [
--- a/subgraph/package.json
+++ b/subgraph/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vetro-app-subgraph",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "license": "MIT",
   "scripts": {
     "prebuild": "npm run codegen",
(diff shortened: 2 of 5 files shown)


ArturDolzan (approved):

gabmontes (changes requested):
Good change — minimal, well-tested, and the backfill-on-resync approach is solid.

One convention issue: `mockVaultCalls` now takes 3 parameters. Project conventions require an object for functions with 2+ params. Worth fixing since we're already touching this function.

Also, the checklist item "Environment variables set in CI, or N/A" is unchecked — no new env vars were added here, so it should just be ticked as N/A.

gndelia:
> Good change — minimal, well-tested, and the backfill-on-resync approach is solid.
> 
> One convention issue: `mockVaultCalls` now takes 3 parameters. Project conventions require an object for functions with 2+ params. Worth fixing since we're already touching this function.
> 
> Also, the checklist item "Environment variables set in CI, or N/A" is unchecked — no new env vars were added here, so it should just be ticked as N/A.

The changes are in AssemblyScript, inline objects are not supported..

gabmontes (approved):

PR #455, opened 2026-06-04
Fix divide-by-zero in Earn subgraph

### LINKED ISSUE #454: Earn subgraph: divide-by-zero aborts indexing (convertToAssets on empty vault + transfer cost-basis)
### Description

The Earn subgraph (`subgraph/src/earn.ts`) aborts indexing with **divide-by-zero** errors. There are two distinct sites where this can happen.

**Site 1 — `vault.convertToAssets(oneShare)` in `handleBlock` (`subgraph/src/earn.ts:55`) — primary, recurring**

`handleBlock` is registered as a **block handler with no filter** (`subgraph.yaml` lines 37-38 and 70-71), so it runs on *every* block from each vault's `startBlock`. It calls `vault.convertToAssets(oneShare)` to compute the share price. An ERC4626-style vault computes this as roughly `shares * totalAssets / totalSupply`. When the vault has **zero total supply** — the window between deployment (`startBlock`) and the first deposit, or any period after all holders fully exit — the contract's internal division panics with a divide-by-zero, the call reverts, and because the code uses the non-`try_` form, the handler aborts. This requires no data divergence; it fires deterministically on every empty-vault block. This is the most likely source of the observed error.

> Note: the `StakingVault` ABI exposes only `convertToAssets`, `decimals`, and `totalAssets` — there is **no `totalSupply`** to pre-check, so the guard must be done via the `try_` variant.

**Site 2 — `.div(fromEntity.shares)` in `handleTransfer` (`subgraph/src/earn.ts:114`) — secondary**

The sender's cost basis is recomputed as `totalCostBasis * newShares / fromEntity.shares`. The guard above it (line 105) only resets to zero when `fromEntity.shares.equals(value)`. If a transfer-out arrives where the sender's tracked `shares` is `0` (or `< value`) while `value > 0`, the `else` branch runs and divides by zero. This requires tracked shares to diverge from the on-chain balance, but it is a real failure mode.

**Safe (no change needed):** `block.timestamp.div(daySeconds)` (`earn.ts:41`) — `daySeconds` is the constant `86400`, never zero.

### Steps to reproduce

Site 1:
1. Index a staking vault from its deployment block while total supply is still zero (no deposits yet), **or** wait until all holders have fully exited.
2. `handleBlock` calls `vault.convertToAssets(oneShare)`, which reverts (internal division by zero on total supply).
3. The handler aborts and indexing fails.

Site 2:
1. A `UserStakingPosition` ends up with `shares = 0` while still existing (e.g. tracked shares diverge from the on-chain balance).
2. A `Transfer` event with that address as `from` and `value > 0` is processed.
3. `fromEntity.shares.equals(value)` is false, the `else` branch runs, and `.div(fromEntity.shares)` divides by zero.

### Environment

- **Component**: `subgraph/` (The Graph / matchstick, AssemblyScript mappings)
- **Affected vaults**: sVUSD (`0x476310E34D2810f7d79C43A74E4D79405bd7a925`), sVetBTC (`0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e`)

### Proposed fix

- **Site 1:** use `vault.try_convertToAssets(oneShare)`; if it reverts, return early and write no `VaultHistory` entry for that block (skip empty-vault blocks).
- **Site 2:** broaden the reset guard from `fromEntity.shares.equals(value)` to `fromEntity.shares.le(value)`, so the proportional `else` branch only runs when `fromEntity.shares > value` (denominator guaranteed non-zero).
- Add matchstick tests: an empty-vault `handleBlock` case (mock `convertToAssets` to revert, assert no `VaultHistory` created) and a zero-share transfer-out case (assert reset to 0, no panic).

A fix implementing the above is in progress on branch `fix-divide-by-zero`.

gndelia:
### Description

Two integer-division sites in `subgraph/src/earn.ts` could divide by zero and abort indexing.

- **`handleBlock` → `convertToAssets` (empty vault):** the block handler runs on every block from each vault's `startBlock`. While a vault has zero total supply (before the first deposit, or after a full exit), the ERC4626 `convertToAssets` call reverts with a divide-by-zero panic, and the non-`try_` call aborted the handler. Fixed by switching to `try_convertToAssets` and skipping the block (no `VaultHistory` row) when it reverts. The `StakingVault` ABI exposes no `totalSupply`, so the `try_` guard is the right approach.
- **`handleTransfer` → proportional cost basis:** the sender's cost basis is recomputed as `totalCostBasis * newShares / fromEntity.shares`, but the reset guard only fired on `shares == value`. A transfer-out from a position whose tracked shares were `0` (or below `value`) fell into the proportional branch and divided by zero. Widened the guard from `.equals(value)` to `.le(value)` so the division only runs when `shares > value` (and `value` is already proven `> 0`), making the denominator provably non-zero.

Added matchstick tests for both cases (empty-vault block, zero-share transfer-out) and bumped the subgraph version to `1.4.1`.

### Screenshots

N/A — subgraph mapping change, no UI.

### Related issue(s)

Closes #454

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraph/package.json
+++ b/subgraph/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vetro-app-subgraph",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "license": "MIT",
   "scripts": {
     "prebuild": "npm run codegen",
--- a/subgraph/src/earn.ts
+++ b/subgraph/src/earn.ts
@@ -41,6 +41,17 @@ export function handleBlock(block: ethereum.Block): void {
   const dayTimestamp = block.timestamp.div(daySeconds).times(daySeconds);
 
   const vaultAddress = dataSource.address();
+
+  const vault = StakingVault.bind(vaultAddress);
+  const decimals = vault.decimals();
+  const oneShare = BigInt.fromI32(10).pow(<u8>decimals);
+  const shareValueResult = vault.try_convertToAssets(oneShare);
+  // Empty vault (zero total supply): convertToAssets reverts with a divide-by-zero
+  // panic. Skip recording history for this block rather than aborting the handler.
+  if (shareValueResult.reverted) {
+    return;
+  }
+
   const id = buildId(vaultAddress, dayTimestamp.toString());
   let entity = VaultHistory.load(id);
   if (entity == null) {
@@ -49,11 +60,7 @@ export function handleBlock(block: ethereum.Block): void {
     entity.stakingVaultAddress = vaultAddress;
   }
 
-  const vault = StakingVault.bind(vaultAddress);
-  const decimals = vault.decimals();
-  const oneShare = BigInt.fromI32(10).pow(<u8>decimals);
-  const shareValue = vault.convertToAssets(oneShare);
-  entity.shareValue = shareValue;
+  entity.shareValue = shareValueResult.value;
   entity.totalAssets = vault.totalAssets();
 
   entity.save();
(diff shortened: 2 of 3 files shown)


gabmontes (approved):
uACK

ArturDolzan (approved):

PR #488, opened 2026-06-18
Add live current value to share-value-history endpoint

### LINKED ISSUE #486: Peg Stability chart lags up to ~24h: append live current value and lower edge cache to 5m

The Peg Stability chart on the Analytics page (`web/src/pages/analytics/index.tsx` → `PegStabilityCard`) shows only historical daily snapshots and never a live/current share value. We want to:

1. **Append a "current" datapoint** to the share-value-history series returned by the API. In parallel with the subgraph query, read the current share value on-chain and append it as the last point once both have resolved.
2. **Reduce the API edge cache** for that endpoint from `max-age=3600` (1h) down to `max-age=300` (5 min), matching the frontend React Query `staleTime`.

Today the chart can lag reality by up to ~24 hours and never reflects the in-progress day.

Data path: subgraph `VaultHistory` → API `GET /variable-stake/share-value-history/:stakingVaultAddress/:period` → `useShareValueHistory` → Victory line chart. Nothing in that path injects a "now" value (`web/src/pages/analytics/components/pegStabilityCard.tsx` plots `chartData` as-is; `web/src/hooks/useShareValueHistory.ts` maps API entries directly).

**Why it lags:**

- The subgraph block handler (`subgraph/src/earn.ts` → `handleBlock`) is a polling handler running ~hourly (`subgraph/subgraph.yaml`, `every: 300` blocks ≈ 1h) and records **at most one `VaultHistory` row per UTC day** (it returns early if today's row already exists).
- Each row is **labeled with the previous day's 00:00 UTC timestamp** but stores the share value read at the current block. So the newest point is x-labeled "yesterday" and holds the value as of ~the start of today (the last UTC-midnight snapshot). There is **no point for the in-progress day**.

**Latency stack for a sudden peg change:**

| Layer | Delay | Source |
|---|---|---|
| Daily snapshot recording | 0–~25h (next UTC midnight + first hourly poll) | `subgraph/src/earn.ts`, `subgraph/subgraph.yaml` |
| API edge cache (`max-age=3600`) | up to 1h | `api/src/index.ts` |
| Frontend React Query `staleTime` | up to 5 min | `web/src/hooks/useShareValueHistory.ts` |

Worst case ≈ 24–25h; the daily-snapshot granularity dominates. The cache and `staleTime` layers can't make data fresher than the daily snapshot.

### Proposed implementation

In `api/src/share-value-history.ts` (`getShareValueHistory`):

- In parallel with the subgraph pagination query, read the **current** share value for the staking vault on-chain. Mirror the subgraph's computation in `handleBlock`: `convertToAssets(oneShare)` then `formatUnits(value, 18)`. Stamp it with the current timestamp.
- Reuse the existing mainnet client (`api/src/mainnet-client.ts`) and StakingVault helpers/ABI (`api/src/staking-vault.ts`) for the read.
- Append the live value as the final datapoint once both resolve, so the chart's last point reflects "now."
- Handle the empty-vault case: `convertToAssets` reverts on zero total supply (the subgraph handler skips this case) — on revert, return the historical series without the appended point.

In `api/src/index.ts`:

- Reduce the cache on the `/variable-stake/share-value-history/...` route from `max-age=3600` to `max-age=300`, matching the frontend `staleTime` of 5 min.

Per project conventions, update `api/README.md` in the same change since this alters the endpoint's response shape (an extra trailing live datapoint).

gndelia:
### Description

The Peg Stability chart on the Analytics page is fed by `GET /variable-stake/share-value-history/:stakingVaultAddress/:period`, which previously returned only the subgraph's daily `VaultHistory` snapshots. The subgraph records one entry per UTC day, so the chart could lag the real share value by up to ~24h and never reflected the in-progress day.

This appends a live datapoint read on-chain via `convertToAssets` (one whole share), run concurrently with the subgraph query, so the series' last value tracks the current exchange rate. If the live read fails (e.g. an empty vault, where `convertToAssets` reverts), the series falls back to the subgraph data alone.

Both the stored and live values are now scaled by the underlying asset's decimals (read on-chain via `asset()` → `decimals()`) instead of a hardcoded `18`, since `convertToAssets` returns asset-denominated amounts.

Finally, the endpoint's edge cache is lowered from 1h to 5m to match the frontend's React Query `staleTime`, so the cache can't serve data staler than the client expects.

### Screenshots

Yesterday, there was an increase in value in `sVUSD` measured in `VUSD`

The snapshot did not impacted yet on the subgraph, per the issue description

Without this change

image

With this change

image

### Related issue(s)

Closes #486

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/api/src/share-value-history.ts
+++ b/api/src/share-value-history.ts
@@ -1,21 +1,80 @@
 import { type Address, formatUnits } from "viem";
+import { decimals } from "viem-erc20/actions";
+import { asset, convertToAssets } from "viem-erc4626/actions";
 
+import { createMainnetClient } from "./mainnet-client.ts";
 import { paginateSubgraphQuery } from "./paginate-subgraph-query.ts";
 import { getPeriodStart } from "./vault-history-period.ts";
 
 type Row = { shareValue: string; timestamp: string };
 
+type Client = ReturnType<typeof createMainnetClient>;
+
+/**
+ * Read the vault's current share value live from the chain: the assets returned
+ * for one whole share (10^shareDecimals), in the asset's base units. The
+ * subgraph only writes a VaultHistory entry once per UTC day, so its latest
+ * point can lag the actual share value by up to ~24h; appending a live point
+ * keeps the series' last value in sync with the on-chain rate. Returns null (so
+ * the caller falls back to the subgraph series alone) if the read fails — which
+ * also covers the empty vault case, where convertToAssets reverts on zero total
+ * supply. The caller scales the returned value by the asset decimals.
+ */
+async function getLiveShareValue({
+  client,
+  stakingVaultAddress,
+}: {
+  client: Client;
+  stakingVaultAddress: Address;
+}) {
+  try {
+    const shareDecimals = await decimals(client, {
+      address: stakingVaultAddress,
+    });
+    return await convertToAssets(client, {
+      address: stakingVaultAddress,
+      // one whole share
+      shares: 10n ** BigInt(shareDecimals),
+    });
+  } catch (error) {
+    console.warn(
+      `Failed to read live share value for ${stakingVaultAddress}: ${error.message}`,
+    );
+    return null;
+  }
+}
+
+/**
+ * Read the decimals of the vault's underlying asset. convertToAssets returns a
+ * value denominated in that asset, so both the subgraph's stored share values
+ * and the live read are scaled by these decimals to a human-readable rate.
+ */
+async function getAssetDecimals({
+  client,
+  stakingVaultAddress,
+}: {
+  client: Client;
+  stakingVaultAddress: Address;
+}) {
+  const assetAddress = await asset(client, { address: stakingVaultAddress });
+  return decimals(client, { address: assetAddress });
+}
+
 /**
  * Query the subgraph for the share value history of a staking vault over the
  * given period. Paginates over the subgraph's 100-result page cap so that long
- * windows (e.g. "1y", up to ~366 daily entries) are returned in full.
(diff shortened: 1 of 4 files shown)


api/src/share-value-history.ts
```diff
@@ -38,19 +97,37 @@ export async function getShareValueHistory({
...
+      cursor: {
+        getValue: (row) => row.timestamp,
+        variable: "start",
+      },
+      field: "vaultHistories",
+      query,
+      url,
+      variables: { stakingVault, start },
+    }),
+    getAssetDecimals({ client, stakingVaultAddress }),
```
gabmontes, line 115: If `asset()` or `decimals()` throws (e.g. RPC unavailable), the whole `Promise.all` rejects and the endpoint returns 500 — discarding perfectly good subgraph data. Unlike `getLiveShareValue`, this has no error guard. Wrapping `getAssetDecimals` in a try/catch falling back to `18` would match pre-PR resilience.

— Auto Gabi

gndelia: `assetDecimals` scales every subgraph row here, not just the live point — this PR replaced the hardcoded `18` with this read. Falling back to `18` on an RPC failure would silently mis-scale all historical rows for any non-18-decimals asset, so we prefer to fail loudly rather than return wrong values. The live read is already guarded (`getLiveShareValue` returns `null` → live point skipped), so a live-read outage still returns the subgraph series; only the decimals read — required to format *any* point — is intentionally a hard dependency. Keeping as-is.

ArturDolzan (approved):

gabmontes (approved):

PR #495, opened 2026-06-19
Fix borrow endpoints after Morpho schema change

### LINKED ISSUE #494: Pool size in Borrow page is not loading
### Description

The pool size is not listing anything on the borrow page

Image

Pool size is not listing this information

### Steps to reproduce

1. Go to the borrow page, see the markets details
2. Pool size should contain information, but renders `-`

I did some quickly troubleshooting, and the endpoint `/borrow/{marketId}/collateral-assets` is being called and returns 404. I think this endpoint should return that data

gndelia:
### Description

The Borrow page "Pool size" rendered `-` because `GET /borrow/{marketId}/collateral-assets` returned 404.

Morpho changed its public GraphQL schema: the `marketByUniqueKey(uniqueKey: ...)` query was removed and replaced by `marketById(marketId: ..., chainId: ...)`, where `chainId` is now required and the `uniqueKey` field was renamed to `marketId`. All three queries in `api/src/morpho.ts` still used the removed field, so each threw — and the borrow route's validation step caught the error and returned 404. This silently broke both `/collateral-assets` (reported) and `/apr-history/:period` (same cause).

The fix switches all three queries to `marketById` and passes the mainnet chain id (these markets live on Ethereum mainnet), using the existing `viem/chains` `mainnet.id` convention.

### Screenshots

image

### Related issue(s)

Closes #494

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/api/src/morpho.ts
+++ b/api/src/morpho.ts
@@ -1,6 +1,10 @@
+import { mainnet } from "viem/chains";
+
 import * as graphql from "./graphql.ts";
 
 const morphoApiUrl = "https://api.morpho.org/graphql";
+
+const chainId = mainnet.id;
 
 export async function getLoanAssetAddress({
   marketId,
@@ -8,24 +12,25 @@ export async function getLoanAssetAddress({
   marketId: string;
 }): Promise<string> {
   const query = `
-    query ($marketId: String!) {
-      marketByUniqueKey(uniqueKey: $marketId) {
+    query ($chainId: Int!, $marketId: String!) {
+      marketById(marketId: $marketId, chainId: $chainId) {
         loanAsset {
           address
         }
       }
     }`;
   const variables = {
+    chainId,
     marketId,
   };
-  const { marketByUniqueKey } = await graphql.runQuery<{
-    marketByUniqueKey: {
+  const { marketById } = await graphql.runQuery<{
+    marketById: {
       loanAsset: {
         address: string;
       };
     };
   }>(morphoApiUrl, query, variables);
-  return marketByUniqueKey.loanAsset.address;
+  return marketById.loanAsset.address;
 }
 
 export async function getHistoricalBorrowApy({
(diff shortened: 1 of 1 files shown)


api/src/morpho.ts
```diff
@@ -1,31 +1,36 @@
...
-    marketByUniqueKey: {
+  const { marketById } = await graphql.runQuery<{
+    marketById: {
       loanAsset: {
         address: string;
       };
     };
   }>(morphoApiUrl, query, variables);
-  return marketByUniqueKey.loanAsset.address;
+  return marketById.loanAsset.address;
```
gabmontes, line 33: `marketById` should be typed as `| null` and guarded before access. With the new `marketById` query, Morpho returns `{ "marketById": null }` (not a GraphQL error) when the market doesn't exist. Without a null check, this propagates as a `TypeError: Cannot read properties of null (reading 'loanAsset')`, which leaks a JS implementation detail into the client-facing error response (`"Invalid market id: Cannot read properties of null..."` instead of something meaningful).

Suggested fix:
```ts
const { marketById } = await graphql.runQuery<{
  marketById: {
    loanAsset: {
      address: string;
    };
  } | null;
}>(morphoApiUrl, query, variables);
if (marketById === null) {
  throw new Error("Market not found");
}
return marketById.loanAsset.address;
```

— Gabi 🤖

gndelia: I just tested it in the Morpho playground and if the market does not exists, an error is returned

```json
{
  "errors": [
    {
      "message": "No results matching given parameters",
      "status": "NOT_FOUND",
      "extensions": {
        "description": "cannot find market config for chainId=1 marketId=0x55609be688a4d96e715bfe39969133bd4e7f83db4f77bb06216109189a11f2e6"
      }
    }
  ],
  "data": null,
  "extensions": {
    "complexity": 30,
    "maximumComplexity": 1000000
  }
}
````

you can check it here https://api.morpho.org/graphql

gabmontes (approved):

ArturDolzan (approved):

PR #522, opened 2026-06-24
Add Earn deposit e2e test with shared helpers

gndelia:
### Description

Adds an end-to-end test for the Earn page stake flow: deposit VUSD into the sVUSD ERC-4626 vault and verify the result on-chain and in the UI.

The test connects the mock wallet (silent EIP-6963 auto-connect), navigates to Earn, opens the VUSD pool's stake drawer, and runs the two-tx "Approve and deposit" path. It then asserts three things: the VUSD balance dropped by exactly the deposited amount; the sVUSD minted (read straight off the deposit tx's ERC-4626 `Deposit` event) matches the sVUSD balance increase; and the staked position surfaces on the Earn page via the on-chain-derived "From 1 pool" badge and its tooltip.

To read the exact transaction the mock wallet sent, this adds a `walletTxHashes` fixture backed by a recording `http` transport that taps `eth_sendRawTransaction`. The wallet executes writes in Node, so the browser never makes that call and Playwright page routing can't observe it — tapping the wallet's own transport is the reliable way to learn which tx it sent.

It also extracts the shared `getMainnetToken` helper and a new `waitForBalance` polling helper into `web/e2e/helpers.ts`, and refactors `swap.spec.ts` to use them (removing the duplicated `getMainnetToken` and the repeated inline `expect.poll(balanceOf …)` blocks).

### Screenshots

https://github.com/user-attachments/assets/757de556-178c-4768-ac42-dbe8deb3a20a

### Related issue(s)

N/A

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/e2e/earn.spec.ts
+++ b/web/e2e/earn.spec.ts
@@ -1,0 +1,144 @@
+import { TEST_ADDRESS } from "@hemilabs/anvil-fork-setup/utils";
+import { expect } from "@playwright/test";
+import { parseAbiItem, parseEventLogs, parseUnits } from "viem";
+import { getTransactionReceipt } from "viem/actions";
+import { balanceOf } from "viem-erc20/actions";
+
+import { sVusdAddress } from "../../packages/earn/src/stakingVaultAddresses.ts";
+
+import { createEthereumClient } from "./anvil";
+import { test } from "./fixtures/wallet";
+import { getMainnetToken, waitForBalance } from "./helpers";
+
+const vusd = getMainnetToken("VUSD");
+const DEPOSIT_DISPLAY = "5";
+const DEPOSIT_AMOUNT = parseUnits(DEPOSIT_DISPLAY, vusd.decimals);
+
+// The VUSD pool is the sVUSD ERC-4626 vault. A deposit emits the standard
+// ERC-4626 Deposit(sender, owner, assets, shares); `shares` is exactly the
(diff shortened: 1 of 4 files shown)


web/e2e/earn.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+  });
+  await expect(submitButton).toBeEnabled();
+  await submitButton.dispatchEvent("click");
+
+  // The success toast renders once both txs confirm on the fork (60s budget to
+  // cover the two sequential signatures + confirmations).
+  await expect(page.getByText("Stake deposit confirmed")).toBeVisible({
+    timeout: 60_000,
+  });
+
```
gabmontes, line 100: The `expect(depositTxHash).toBeDefined()` check only guards against an empty array. If an extra tx is ever sent (e.g. a permit or a second approval), `.at(-1)` will silently pick the wrong hash and the failure will land further down as `Expected array to have length 1, but has length 0` from `parseEventLogs` — not immediately obvious. Consider adding `expect(walletTxHashes).toHaveLength(2)` before `.at(-1)` so the failure points directly at the unexpected tx count.

— Gabi 🤖

gndelia: no problem, we'll most likely always want the last tx. We'll revisit if it fails

gabmontes (approved):

ArturDolzan (approved):

PR #539, opened 2026-06-25
web: allow inline scripts in dev for Vite preamble

gndelia:
### Description

[PR #535](https://github.com/vetro-protocol/vetro-monorepo/pull/535) tightened the worker CSP and dropped `'unsafe-inline'` from `script-src`. In dev, Vite injects an inline React Fast Refresh preamble script, so the stricter policy blocked it and every component threw `@vitejs/plugin-react can't detect preamble`, breaking local dev.

This re-adds `'unsafe-inline'` to `script-src` only when [`import.meta.env.DEV`](https://vite.dev/guide/env-and-mode) is set, so production keeps the strict policy.

### Checklist

- [ ] Manual testing passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/src/index.ts
+++ b/web/src/index.ts
@@ -70,6 +70,16 @@ const connectSrc = [
 
 const workerSrc = import.meta.env.VITE_SENTRY_DSN ? "blob:" : "'none'";
 
+// In dev, Vite injects an inline React Fast Refresh preamble script, so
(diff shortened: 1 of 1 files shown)


joshuasing (approved):
LGTM

gabmontes (approved):

PR #542, opened 2026-06-25
web: add nonce to Content-Security-Policy

joshuasing:
### Description

Add nonce to `Content-Security-Policy`.

The nonce in script-src is redundant with 'self' for our own bundled scripts, but Cloudflare's JavaScript Detections parses the CSP response header and reuses this nonce for the inline scripts it injects.
https://developers.cloudflare.com/cloudflare-challenges/challenge-types/javascript-detections/#if-you-have-a-content-security-policy-csp

If we later need to add inline scripts, this can be injected for those.

### Related issue(s)

Closes #
Related to #

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/index.ts
+++ b/web/src/index.ts
@@ -157,7 +157,7 @@ const buildCsp = (nonce: string) =>
     // (bot management / Turnstile) render in iframes.
     `frame-src 'self' ${walletFrameSrc} https://challenges.cloudflare.com`,
     `img-src 'self' data: https://hemilabs.github.io ${walletImgSrc}`,
-    `script-src ${scriptSrc} nonce-${nonce}`,
+    `script-src ${scriptSrc} 'nonce-${nonce}'`,
     // Tailwind v4 injects styles via a <style> tag, so 'unsafe-inline' is needed.
     "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
     `worker-src ${workerSrc}`,
--- a/web/src/index.ts
+++ b/web/src/index.ts
@@ -132,23 +132,37 @@ const permissionsPolicy = [
   .map((feature) => `${feature}=()`)
   .join(", ");
 
-const csp = [
-  "base-uri 'none'",
-  `connect-src ${connectSrc}`,
-  "default-src 'none'",
-  "font-src 'self' https://fonts.gstatic.com",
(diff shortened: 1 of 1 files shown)


gndelia (dismissed):

web/src/index.ts
```diff
@@ -132,23 +132,37 @@ const permissionsPolicy = [
...
+    `connect-src ${connectSrc}`,
+    "default-src 'none'",
+    "font-src 'self' https://fonts.gstatic.com",
+    "form-action 'none'",
+    "frame-ancestors 'none'",
+    // WalletConnect verify + Coinbase smart wallet + Cloudflare challenges
+    // (bot management / Turnstile) render in iframes.
+    `frame-src 'self' ${walletFrameSrc} https://challenges.cloudflare.com`,
+    `img-src 'self' data: https://hemilabs.github.io ${walletImgSrc}`,
+    `script-src ${scriptSrc} nonce-${nonce}`,
```
gndelia:
```suggestion
    `script-src ${scriptSrc} 'nonce-${nonce}'`,
```

FYI unchecked, but Claude says this should be quoted. Worth checking! (Leaving my approval in case the suggestion is incorrect)

joshuasing: Ah good catch, it looks like it should be quoted. Thanks!
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP

joshuasing: I think I actually had it quoted and removed it when merging changes :')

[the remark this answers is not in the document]
joshuasing: Fixed!

gndelia (changes requested):
Now that you fixed the nonce, it locally breaks with `pnpm run dev`. Test it in your local

joshuasing:
Whoops, tested with `pnpm preview` instead of HMR again, apologies.

Looks like `nonce` overrides `unsafe-inline` in modern browsers, so this effectively overrides your fix from earlier. I am going to make it exclude `nonce` when running in dev mode.

joshuasing:
Thanks, fixed and tested with both `pnpm dev` and `pnpm preview`.

gndelia (approved):

ArturDolzan (approved):
uACK

Issue #474, opened 2026-06-15
Add internal dashboard to track metrics


gndelia:
### Feature or change description

Not a public dashboard - we could use CF protecion with hemilabs accounts

To begin with, we. should track liquidity of Vetro tokens on Curve Pools

Other metrics are also accepted

mitakash:
https://docs.google.com/document/d/1v0P3vKjH6iJmrXYa_b6aGrDnsPJBv3RBExrUaS8zfrM/edit?usp=sharing
Also all addresses are here --->

https://github.com/vetro-protocol/vetro-contracts/issues/111
@gndelia

gndelia:
@mitakash what would you like to track from each token?

I was thinking something like

- TVL of each token per pool (live value), and maybe a historic chart of locked
- distribution of tokens in DEX (so how much % of, say, VUSD is in each pool, given all add up to 100%). We calculate of all VUSD/VETBTC, etc that is locked, what's the proportion per pool
- APY per pool

Anything else you consider relevant?

gndelia:
Posting here so it's more accesible

| Pool  | Pool Address | Gauge Address | Curve deposit link |
  | ------------- | ------------- | ------------- | ------------- |
  | VUSD/crvUSD | 0xafba5800252530ce71b03ba2bca2dd5ae44a7f3d | 0x737e7700e03a8c451c9b72103554a40760f1b57a |  https://www.curve.finance/dex/ethereum/pools/0xafba5800252530ce71b03ba2bca2dd5ae44a7f3d/swap |
  | crvUSD/sVUSD | 0x659b7b5dd7936bf2f2d198a87c1583049d1d91d3 | 0x543c88e9b7bd90a4c14a31bdbf41dd509ab666ed |  https://www.curve.finance/dex/ethereum/pools/0x659b7b5dd7936bf2f2d198a87c1583049d1d91d3/swap |
  | VUSD/hemiBTC | 0xc4ff0e84ec219581092e5660c74a2da8de6aee4f | 0x5B933Fc47431795a0c2eb8D183BEAE9722e1fD83 |  https://www.curve.finance/dex/ethereum/pools/0xc4ff0e84ec219581092e5660c74a2da8de6aee4f/deposit |
[...]

PR #560, opened 2026-06-30
Add DEX per-pool detail page to dashboard

gndelia:
### Description

Adds a per-pool detail page to the internal dashboard at `/dex/:poolAddress`, reachable from the DEX list. It shows TVL for any pool and, for Curve pools, 24h volume, fees, APY and liquidity utilization (24h volume / TVL). Every pool also gets a per-coin liquidity breakdown (balance, value, price, share), an exchange-rate card that surfaces peg drift for stable pairs, Curve gauge emissions (gauge weight, est. CRV/day, CRV APY), and the pool/gauge/LP-token/coin addresses with explorer links.

Also drops a few now-unused type/option exports flagged by knip (`curveApi` `Curve*` types and web's `tokenPricesOptions`).

### Screenshots

https://github.com/user-attachments/assets/cb5286fa-f48f-4e9f-8e01-14a6a4cc61a7

### Related issue(s)

Related to #474

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/internal-dashboard/src/fetchers/fetchGaugeEmissions.ts
+++ b/internal-dashboard/src/fetchers/fetchGaugeEmissions.ts
@@ -17,7 +17,14 @@ export const fetchGaugeEmissions = async function (): Promise<
     }
     const inflationRate = Number(gauge.inflationRate) / 1e18;
     const relativeWeight = Number(gauge.relativeWeight) / 1e18;
-    emissions[gauge.pool.toLowerCase()] = {
+    const key = gauge.pool.toLowerCase();
+    // A pool can expose several gauges (old/killed ones linger). Keep the one
+    // emitting the most CRV so a dead gauge can't zero out an active pool.
+    const existing = emissions[key];
+    if (existing && existing.inflationRate >= inflationRate) {
+      continue;
+    }
+    emissions[key] = {
       estCrvPerDay: inflationRate * relativeWeight * SECONDS_PER_DAY,
       inflationRate,
       relativeWeight,
--- a/internal-dashboard/src/fetchers/fetchGaugeEmissions.ts
+++ b/internal-dashboard/src/fetchers/fetchGaugeEmissions.ts
@@ -1,0 +1,27 @@
+import { fetchGauges } from "../lib/curveApi";
+import { type GaugeEmission } from "../lib/types";
+
+const SECONDS_PER_DAY = 86_400;
+
+// Normalizes every gauge's emission data, keyed by its pool address (lowercased)
+// for cheap lookup from a pool detail view.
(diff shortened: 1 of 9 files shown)


internal-dashboard/src/pages/dexPool.tsx
```diff
@@ -0,0 +1,352 @@
...
+    return (
+      <StateMessage>
+        Couldn&apos;t load pool data. Try again later.
+      </StateMessage>
+    );
+  }
+
+  const pool = pools?.find(
+    (candidate) => candidate.id.toLowerCase() === poolAddress?.toLowerCase(),
+  );
```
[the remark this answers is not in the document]
gndelia: Fixed in [c4660e5](https://github.com/vetro-protocol/vetro-monorepo/pull/560/commits/c4660e5f). The route param and local variable are now `poolId`, matching that the page routes by `TrackedPool.id` (composite for range views), not a raw address.

internal-dashboard/src/fetchers/fetchGaugeEmissions.ts
```diff
@@ -0,0 +1,27 @@
...
+  const gauges = await fetchGauges();
+
+  const emissions: Record<string, GaugeEmission> = {};
+  for (const gauge of gauges) {
+    if (!gauge.pool) {
+      continue;
+    }
+    const inflationRate = Number(gauge.inflationRate) / 1e18;
+    const relativeWeight = Number(gauge.relativeWeight) / 1e18;
+    emissions[gauge.pool.toLowerCase()] = {
```
gabmontes: 77 Curve pools currently have multiple gauges for the same `swap` address. Since the loop overwrites the map entry on each iteration, the last gauge processed wins — and in 39 of those cases the active gauge (nonzero `inflation_rate`) appears before a dead one (zero inflation), so emissions silently zero out.

Filtering killed gauges in `fetchGauges` (the API exposes `is_killed`) or preferring the gauge with the highest `inflation_rate` here would fix it.

— Gabi 🤖

gndelia: Fixed in [e659d6f](https://github.com/vetro-protocol/vetro-monorepo/pull/560/commits/e659d6f). Confirmed against the live `getAllGauges` API: 77 pools have multiple gauges and 39 have an active gauge before a dead one — matching your numbers. Filtering `is_killed` alone wasn't enough (67 pools still have multiple *live* gauges where only one emits), so the loop now keeps the gauge with the highest `inflationRate` per pool.

gabmontes (approved):

ArturDolzan (approved):

PR #588, opened 2026-07-03
web: use matching API deployment for preview URLs

### LINKED ISSUE #496: FE deployed in Preview URLs uses the staging API instead of the API from the preview
### Description

CF deploys 4 things as preview URLs - the relevant being the Vetro FE and its API. However, the Frontend uses the staging API instead of the version published in this PR. It would be ideal the preview FE uses the previeW API

Use case: A PR fixes something in the API, and I can't test it in the preview URL because the FE doesn't use it (of course I can use postman, plus it was verified locally, but still)

### Proposed fix

get predictable urls built with the commit and use that as an env variable

joshuasing:
### Description

Preview deployments of the web worker now connects to the matching API worker preview deployment instead of the shared staging API. The API origin is derived from the web preview's own hostname (`<branch>-vetro-web-staging.<subdomain>` -> `<branch>-vetro-api-staging.<subdomain>`) and used for both the client's API calls (`getVetroApiUrl`) and the worker's CSP `connect-src`.

This applies for preview deployments only. Derivation is gated on `VITE_DEPLOY_ENV === "preview"`. Vite should dead-code eliminate this from the production bundle so this can never run in prod.

The frontend will only use the preview API URL for branch/alias preview deployments. The "Commit Preview URL"s use a separate alias/hash per worker, preventing us from determining the API URL for the web URL.

Also add all `VITE_*` vars to `vite-env.d.ts`.

### Related issue(s)

Closes #496
Related to #

### Checklist

- [x] Manual testing passed.


web/src/utils/url.ts
```diff
@@ -1,5 +1,35 @@
 export const isRelativeUrl = (url: string) => url.startsWith("/");
 
+// Maps a web staging preview host to its matching API preview. The preview host
+// suffixes come from the preview build env. Gated to the preview build so it's
+// dead-code eliminated everywhere else.
+export function deriveApiOrigin(hostname: string): string | undefined {
```
gndelia: I think we can remove this as it can be inferred

```suggestion
export function deriveApiOrigin(hostname: string) {
```

gndelia (commented):
I left a minor suggestion, but looks good. I also tested it and Web is using the Preview API 🎉

web/README.md
```diff
@@ -8,14 +8,16 @@ Built with React, Viem, Wagmi, and Tailwind CSS.
...
+| `VITE_STAGING_API_PREVIEW_HOST`  | No       | API preview host suffix (`<api-worker>.<subdomain>`) used to build the matching API preview URL. Only read when `VITE_DEPLOY_ENV=staging`.                                                                                                                                                                      |
+| `VITE_STAGING_WEB_PREVIEW_HOST`  | No       | Web preview host suffix (`<web-worker>.<subdomain>`) matched to detect a staging preview and extract its alias. Only read when `VITE_DEPLOY_ENV=staging`.                                                                                                                                                       |
+| `VITE_WALLET_CONNECT_PROJECT_ID` | No       | WalletConnect project ID.                                                                                                                                                                                                                                                                                       |
```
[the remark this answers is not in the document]
joshuasing: This would be an unrelated change, better suited to a separate PR.

gndelia (commented):
I think this branch needs to be rebased, and `web/src/hooks/useSubmitContactForm.ts` needs to be included to work with `getVetroApiUrl()`

web/src/vite-env.d.ts
```diff
@@ -1 +1,13 @@
 /// <reference types="vite/client" />
+
+interface ImportMetaEnv {
+  readonly VITE_CONTACT_FORM_ENABLED?: string;
```
gndelia, line 4: No longer needed - it was removed in  https://github.com/vetro-protocol/vetro-monorepo/pull/623
```suggestion
```

gndelia (changes requested):

web/src/utils/url.ts
```diff
@@ -1,5 +1,35 @@
...
+  }
+  const previewAlias = hostname.slice(0, -webHost.length - 1);
+  // Only branch aliases are shared across the web/api workers; per-version
+  // preview prefixes are an 8 hex char token, so skip those.
+  const isBranchAlias =
+    /^[a-z0-9-]+$/.test(previewAlias) && !/^[0-9a-f]{8}$/.test(previewAlias);
+  return isBranchAlias ? `https://${previewAlias}-${apiHost}` : undefined;
+}
+
+export function getVetroApiUrl(): string | undefined {
```
gndelia, line 23: I think this can be removed - it's redundant
```suggestion
export function getVetroApiUrl() {
```

gndelia:
@joshuasing conflicts ! Can we get this solved? It's close, but it's missing a rebase + a few bits!

PR #617, opened 2026-07-14
Block sanctioned addresses

### LINKED ISSUE #549: Geoblock sanctioned countries
Apply geoblocking/geofencing with an approach similar to Avant.

gabmontes:
### Description

This PR adds logic to detect and block OFAC sanctioned addresses, on top of sanctioned and restricted countries. It will display a full-screen modal right after wallet connection if the address is sanctioned.

#### Details

This pull request introduces a wallet-based sanctions check and restriction UI to the application, ensuring users with sanctioned addresses are blocked from using the app and shown an appropriate message. It also refactors background illustration code for reuse and updates translation strings for consistency.

**Sanctions enforcement and restriction UI:**

- Added a new `useIsSanctioned` hook that queries an on-chain sanctions oracle to determine if the connected wallet address is sanctioned, blocking access if so.
- Introduced the `AddressRestriction` component, which overlays the app with a restriction dialog when a sanctioned address is detected, including a disconnect wallet action. This component is now rendered at the app root. [[1]](diffhunk://#diff-b865019bcbba92ea86f43de9775b7297fa810158e3654828a6c5373cdaa1815dR1-R54) [[2]](diffhunk://#diff-20da45cb8443d5036aea448ba7f0379ca304c43e1136ca82d9d43d731a37311eR11) [[3]](diffhunk://#diff-20da45cb8443d5036aea448ba7f0379ca304c43e1136ca82d9d43d731a37311eR24)

**UI component improvements:**

- Created a reusable `StatusMessage` component for displaying alert dialogs with icon, title, description, and action.
- Refactored the globe illustration and background into a new `GlobeIllustration` component, updating `AppLayout` to use it. [[1]](diffhunk://#diff-11bbde8b6777448179439b079a0bcd5e4cfa483e4e514b7d69dcc53f2af608abR1-R10) [[2]](diffhunk://#diff-6854d4f4c26bbf2f807f0523266858d2789b9731bbbb9572e74b40320bcfcb31R4-R5) [[3]](diffhunk://#diff-6854d4f4c26bbf2f807f0523266858d2789b9731bbbb9572e74b40320bcfcb31L29-R31)

**Localization updates:**

- Added and unified translation keys for address and geo restrictions, ensuring consistent messaging in both English and Spanish. [[1]](diffhunk://#diff-0ef18c6cbc913ab585ed021069cf18aadb3777fd596a23b18e90d6afd4680895R3) [[2]](diffhunk://#diff-0ef18c6cbc913ab585ed021069cf18aadb3777fd596a23b18e90d6afd4680895L18) [[3]](diffhunk://#diff-0ef18c6cbc913ab585ed021069cf18aadb3777fd596a23b18e90d6afd4680895R28) [[4]](diffhunk://#diff-6362fcecd648c0e6ceb3f9a11087ba350f1aa30224c7dcac382d2bd7c9cc4f6bR3) [[5]](diffhunk://#diff-6362fcecd648c0e6ceb3f9a11087ba350f1aa30224c7dcac382d2bd7c9cc4f6bL18) [[6]](diffhunk://#diff-6362fcecd648c0e6ceb3f9a11087ba350f1aa30224c7dcac382d2bd7c9cc4f6bR28)
- Updated `GeoRestrictionRibbon` to use the new unified restriction description key.

### Screenshots

https://github.com/user-attachments/assets/5709f102-e443-4a90-bf72-681b8c93771a

### Related issue(s)

Closes #549

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/components/addressRestriction.tsx
+++ b/web/src/components/addressRestriction.tsx
@@ -24,11 +24,11 @@ export function AddressRestriction() {
       className="font-geist fixed inset-0 z-9999 overflow-hidden bg-gray-50"
       role="alertdialog"
     >
-      <div className="relative mx-4 h-full border-x border-gray-200 lg:mx-auto lg:max-w-[1024px]">
-        <div className="relative flex h-[160px] items-center justify-center overflow-hidden lg:h-[152px]">
+      <div className="relative mx-4 h-full border-x border-gray-200 lg:mx-auto lg:max-w-5xl">
+        <div className="relative flex h-40 items-center justify-center overflow-hidden lg:h-[152px]">
           <img
             alt=""
-            className="pointer-events-none absolute top-1/2 left-1/2 hidden h-[240px] w-[1365px] -translate-x-1/2 translate-y-[-20%] opacity-3 lg:block"
+            className="pointer-events-none absolute top-1/2 left-1/2 hidden h-60 w-[1365px] -translate-x-1/2 translate-y-[-20%] opacity-3 lg:block"
             src={vetroLogo}
           />
           <img alt="Vetro" className="relative z-10 h-10" src={vetroLogo} />
@@ -46,7 +46,7 @@ export function AddressRestriction() {
           />
         </div>
       </div>
-      <div className="pointer-events-none fixed bottom-0 left-1/2 w-[calc(100%-2rem)] -translate-x-1/2 lg:max-w-[1024px]">
+      <div className="pointer-events-none fixed bottom-0 left-1/2 w-[calc(100%-2rem)] -translate-x-1/2 lg:max-w-5xl">
         <GlobeIllustration />
       </div>
     </div>
--- a/web/src/components/addressRestriction.tsx
+++ b/web/src/components/addressRestriction.tsx
@@ -1,0 +1,54 @@
+import { useTranslation } from "react-i18next";
+import { useDisconnect } from "wagmi";
+
+import { useIsSanctioned } from "../hooks/useIsSanctioned";
+
+import { Button } from "./base/button";
+import { StatusMessage } from "./base/statusMessage";
+import { GlobeIllustration } from "./globeIllustration";
+import vetroLogo from "./icons/vetroLogo.svg";
+import { WarningCircleIcon } from "./icons/warningCircleIcon";
+
+export function AddressRestriction() {
+  const isSanctioned = useIsSanctioned();
+  const { disconnect } = useDisconnect();
+  const { t } = useTranslation();
+
+  if (!isSanctioned) {
+    return null;
+  }
+
+  return (
+    <div
+      aria-modal="true"
+      className="font-geist fixed inset-0 z-9999 overflow-hidden bg-gray-50"
+      role="alertdialog"
+    >
+      <div className="relative mx-4 h-full border-x border-gray-200 lg:mx-auto lg:max-w-[1024px]">
+        <div className="relative flex h-[160px] items-center justify-center overflow-hidden lg:h-[152px]">
+          <img
+            alt=""
+            className="pointer-events-none absolute top-1/2 left-1/2 hidden h-[240px] w-[1365px] -translate-x-1/2 translate-y-[-20%] opacity-3 lg:block"
+            src={vetroLogo}
+          />
+          <img alt="Vetro" className="relative z-10 h-10" src={vetroLogo} />
+        </div>
+        <div className="flex flex-col items-center justify-center border-y border-gray-200 bg-white py-20">
+          <StatusMessage
+            action={
+              <Button onClick={() => disconnect()} size="xSmall">
+                {t("pages.wallet.disconnect-wallet")}
+              </Button>
+            }
+            description={t("common.restriction-description")}
+            icon={<WarningCircleIcon />}
+            title={t("common.address-restriction-title")}
+          />
+        </div>
+      </div>
+      <div className="pointer-events-none fixed bottom-0 left-1/2 w-[calc(100%-2rem)] -translate-x-1/2 lg:max-w-[1024px]">
+        <GlobeIllustration />
+      </div>
+    </div>
+  );
+}
--- a/web/src/components/base/statusMessage.tsx
+++ b/web/src/components/base/statusMessage.tsx
@@ -11,7 +11,7 @@ export const StatusMessage = ({ action, description, icon, title }: Props) => (
   <div className="flex max-w-60 flex-col items-center gap-3 text-center">
     <div className="size-7">{icon}</div>
     <div className="flex flex-col gap-1">
-      <h5 className="text-h5 text-gray-900">{title}</h5>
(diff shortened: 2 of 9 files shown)


web/src/hooks/useIsSanctioned.ts
```diff
@@ -0,0 +1,53 @@
...
+  // seems likely that RPC errors will also prevent any use of the app so this
+  // is a reasonable tradeoff.
+  const { data: isSanctioned = false } = useQuery({
+    enabled: isConnected && !!address && !!client,
+    queryFn: () => fetchIsSanctioned({ address: address!, client: client! }),
+    // The oracle lives on mainnet and the result is chain-agnostic, so
+    // address alone is a sufficient cache key.
+    queryKey: ["is-sanctioned", address],
+    retry: false,
+  });
```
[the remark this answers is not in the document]
gabmontes: Ok. Done.

dvnahuel:
Maybe in this particular case it would be better to just have a white screen saying that the address is sanctioned and you can't use the app with a CTA that would open the wallet connection drawer. Right now the flow has the same content as the geo-block one but we're actually validating the wallet address. thoughts? @gabmontes

web/src/components/addressRestriction.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (!isSanctioned) {
+    return null;
+  }
+
+  return (
+    <div
+      aria-modal="true"
+      className="font-geist fixed inset-0 z-9999 overflow-hidden bg-gray-50"
+      role="alertdialog"
+    >
```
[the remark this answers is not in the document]
gabmontes: Tailwind class is properly generated. Manually checked. And adding `aria-label` seems unnecesary.

gabmontes:
@dvnahuel very similar but the message now points to the user address. I think the flow is correct: The user opens the app, connects the wallet, the address is checked; if sanctioned, full-screen modal with "disconnect" button.

web/src/components/addressRestriction.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <div className="relative mx-4 h-full border-x border-gray-200 lg:mx-auto lg:max-w-[1024px]">
```
ArturDolzan: Any reason for the arbitrary `lg:max-w-[1024px]` here instead of `lg:max-w-5xl`? They resolve to the same 1024px and we already use the `5xl` token for the page width elsewhere, so it'd read a bit more consistent. Not blocking though.

web/src/components/addressRestriction.tsx
```diff
@@ -0,0 +1,54 @@
...
+                {t("pages.wallet.disconnect-wallet")}
+              </Button>
+            }
+            description={t("common.restriction-description")}
+            icon={<WarningCircleIcon />}
+            title={t("common.address-restriction-title")}
+          />
+        </div>
+      </div>
+      <div className="pointer-events-none fixed bottom-0 left-1/2 w-[calc(100%-2rem)] -translate-x-1/2 lg:max-w-[1024px]">
```
ArturDolzan: Same here.

ArturDolzan (commented):
I left two minor comments otherwise LGTM.

web/src/components/addressRestriction.tsx
gabmontes: Fixed!

web/src/components/addressRestriction.tsx
gabmontes: Fixed too.

ArturDolzan (dismissed):

web/src/components/base/statusMessage.tsx
```diff
@@ -0,0 +1,19 @@
...
+  description: string;
+  icon: ReactNode;
+  title: string;
+};
+
+export const StatusMessage = ({ action, description, icon, title }: Props) => (
+  <div className="flex max-w-60 flex-col items-center gap-3 text-center">
+    <div className="size-7">{icon}</div>
+    <div className="flex flex-col gap-1">
+      <h5 className="text-h5 text-gray-900">{title}</h5>
```
gndelia: Redundant - `h5` already defines `text-h5` - see [source](https://github.com/vetro-protocol/vetro-monorepo/blob/d8ec3e4d714f94e705231793aac192dab9d56ff3/web/src/index.css#L184)

```suggestion
      <h5 className="text-gray-900">{title}</h5>
```

gabmontes: Fixed!

web/src/components/addressRestriction.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!isSanctioned) {
+    return null;
+  }
+
+  return (
+    <div
+      aria-modal="true"
+      className="font-geist fixed inset-0 z-9999 overflow-hidden bg-gray-50"
+      role="alertdialog"
```
gndelia, line 25: Maybe it's me, but

Image

this doesn't seem to be a modal to me. IT just seems to be a page

web/src/hooks/useIsSanctioned.ts
```diff
@@ -0,0 +1,56 @@
...
+    // The oracle lives on mainnet and the result is chain-agnostic, so
+    // address alone is a sufficient cache key.
+    queryKey: ["is-sanctioned", address],
+    retry: false,
+    // Sanction status changes very infrequently; avoid unnecessary mainnet RPC
+    // calls on remounts and window refocus.
+    staleTime: 1000 * 60 * 60,
+  });
+
+  return isConnected && isSanctioned;
```
gndelia: This is redundant

```suggestion
  return isSanctioned;
```

if `!isConnected`, then `enabled` is false, so the query does not run, causing `isSanctioned` to be `undefined` and defaulting to `false` per line 43

web/src/components/addressRestriction.tsx
gabmontes: It is a full-screen overlay. What do you suggest?

gndelia: ah, it's ok. Visually it did not look like this - just as a regular page.

web/src/hooks/useIsSanctioned.ts
gabmontes: It is needed on disconnect. If the wallet is sanctioned and the user disconnects, the hook would still return "true" and the full-screen alert will never disappear.

gndelia: How? If the user disconnects, `address` is undefined, so query is disabled. The output of `useQuery` should be `data` as undefined.

It seems there's a bug in your hook then

dvnahuel (approved):
LGTM!

web/src/hooks/useIsSanctioned.ts
gabmontes: Hook simplified and re-tested.

gndelia (approved):

ArturDolzan (approved):
LGTM.

PR #647, opened 2026-07-24
Add Umami analytics for page views and wallets

### LINKED ISSUE #614: Add analytics

Or some option that let us understand how the user uses the app!

gndelia:
### Description

Enables analytics with Umami (issue #614): locale-stripped page views on route change, and explicit EVM wallet connections. The tracker is injected into `index.html` at build time by a Vite plugin, only when both `VITE_ANALYTICS_URL` and `VITE_ANALYTICS_WEBSITE_ID` are set; the worker CSP is widened to allow the tracker origin (`script-src` + `connect-src`).

**These env vars are intentionally not configured yet, so this can be merged and released safely — analytics stays fully disabled (no tracker injected, no tracking component mounted, CSP unchanged) until the vars are set.**

### Screenshots

N/A — no UI changes.

### Related issue(s)

Closes #614

### Checklist

- [x] Manual testing passed.
- [ ] Environment variables set in CI, or N/A. — intentionally deferred; feature ships disabled until configured.

### DIFF
--- a/web/src/components/analyticsTracker.tsx
+++ b/web/src/components/analyticsTracker.tsx
@@ -1,0 +1,39 @@
+import * as Sentry from "@sentry/react";
+import { normalizePath } from "i18n/path";
+import { useCallback, useEffect } from "react";
+import { useLocation } from "react-router";
+import { type Connector, useAccountEffect } from "wagmi";
+
+export function AnalyticsTracker() {
+  const { pathname } = useLocation();
+  const url = normalizePath(pathname);
+
+  useEffect(
+    function trackPageView() {
+      if (url !== "/") {
+        window.umami?.track((props) => ({ ...props, url }));
+      }
+    },
+    [url],
+  );
+
+  useAccountEffect({
+    onConnect: useCallback(function ({
+      connector,
+      isReconnected,
+    }: {
+      connector: Connector;
+      isReconnected: boolean;
+    }) {
+      Sentry.setTag("evm wallet", connector.name);
+      if (!isReconnected) {
+        window.umami?.track("evm connected", { wallet: connector.name });
+      }
+    }, []),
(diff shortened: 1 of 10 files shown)


web/src/index.ts
```diff
@@ -53,12 +53,15 @@ const walletFrameSrc = [
...
+
 const allUrls: (string | undefined)[] = [
   import.meta.env.VITE_PORTAL_API_URL,
   import.meta.env.VITE_VETRO_API_URL,
   ...rpcUrls,
   ...walletConnectUrls,
   import.meta.env.VITE_SENTRY_DSN,
+  import.meta.env.VITE_ANALYTICS_URL,
   "https://cloudflareinsights.com", // Cloudflare Web Beacon analytics.
 ];
```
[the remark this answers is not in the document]
gndelia: not a problem

gabmontes (approved):

web/src/components/analyticsTracker.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useAccountEffect({
+    onConnect: useCallback(function ({
+      connector,
+      isReconnected,
+    }: {
+      connector: Connector;
+      isReconnected: boolean;
+    }) {
+      Sentry.setTag("evm wallet", connector.name);
```
ArturDolzan, line 28: Perhaps worth double-checking this one — I think Sentry doesn't allow spaces in tag keys (only `a-zA-Z0-9_.:-`), so `"evm wallet"` might get dropped on ingest and the wallet name wouldn't actually show up on error reports. Would `evm_wallet` be safer here?

gndelia: While true, Sentry automatically parses it. See how it arrives

image

ArturDolzan (approved):

PR #672, opened 2026-07-30
Minor improvements to the dev fork scripts

gndelia:
### Description

A few small improvements to the `web/scripts` dev-fork tooling, mostly for consistency between the scripts:

- `devFork.ts` now accepts a `--fork-url` (`-f`) flag to override the RPC it forks from, falling back to `VITE_RPC_URL_MAINNET` and then to the default mainnet RPC, as before.
- `fastForwardTime.ts` keeps `--fork-url` as an option, but it's no longer required: it now defaults to the local anvil endpoint (`http://127.0.0.1:8545`), which is what it's pointed at in practice.

More minor script improvements are likely to follow in subsequent PRs.

### Screenshots

N/A

### Related issue(s)

N/A

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/scripts/fastForwardTime.ts
+++ b/web/scripts/fastForwardTime.ts
@@ -48,7 +48,7 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
 
   if (!values.seconds || !/^\d+$/.test(values.seconds)) {
     console.error(
-      "Seconds is invalid. Usage: node web/scripts/fastForwardTime.ts --seconds 120",
+      "Seconds is invalid. Usage: node web/scripts/fastForwardTime.ts --seconds 120 [--fork-url …]",
     );
     process.exit(1);
   }
--- a/web/scripts/fastForwardTime.ts
+++ b/web/scripts/fastForwardTime.ts
@@ -16,10 +16,10 @@ import { mainnet } from "viem/chains";
  * @returns The fork's new block.timestamp (unix seconds) after the jump.
  */
 export async function fastForwardTime({
-  forkUrl,
+  forkUrl = "http://127.0.0.1:8545",
   seconds,
 }: {
-  forkUrl: string;
+  forkUrl?: string;
   seconds: number;
 }) {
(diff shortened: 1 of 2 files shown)


ArturDolzan (dismissed):

web/scripts/fastForwardTime.ts
```diff
@@ -48,14 +48,7 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
...
     console.error(
-      "Seconds is invalid. Usage: node web/scripts/fastForwardTime.ts --fork-url http://127.0.0.1:8545 --seconds 120",
-    );
-    process.exit(1);
-  }
-
-  if (!values["fork-url"]) {
-    console.error(
-      "Fork URL is required. Usage: node web/scripts/fastForwardTime.ts --fork-url http://127.0.0.1:8545 --seconds 120",
+      "Seconds is invalid. Usage: node web/scripts/fastForwardTime.ts --seconds 120",
```
gabmontes: The error message omits `--fork-url`, but the usage comment just above documents it as `[--fork-url …]`. A user hitting this error won't know the flag exists. Suggest aligning them:

```
"Seconds is invalid. Usage: node web/scripts/fastForwardTime.ts --seconds 120 [--fork-url ...]"
```

— Gabi 🤖

gabmontes: If the error is just for `seconds`, that's ok. But in that case, showing a partial "usage" hint may be confusing. Perhaps all errors should either tell the user to call `help` or reuse the output of that command.

– Real Gabi

gndelia: I added the optional `--fork-url` hint here [054a1f3](https://github.com/vetro-protocol/vetro-monorepo/pull/672/commits/054a1f3071c24154e263fbcef60d76d486caca5a)

ArturDolzan (approved):

gabmontes (approved):

PR #676, opened 2026-07-31
Throw when a fork setup transaction reverts

gndelia:
### Description

Small improvement to the `web/scripts/` folder, following up #672.

The scripts that drive the local Anvil fork waited for transaction receipts but never checked `receipt.status`, so a reverted transaction was indistinguishable from a successful one — a couple of them even print a success line right afterwards.

In practice most reverts already surface at send time through anvil's gas estimation, but that's a property of the node rather than of these scripts. It matters because `web/e2e/*.spec.ts` imports several of them as fixtures: a silently reverted setup transaction shows up later as a confusing UI assertion instead of a failed setup step.

This adds a `confirmTransaction` helper in `web/scripts/utils.ts` that throws on a non-success receipt, and routes all 18 call sites across 9 scripts through it. No behavior change while everything succeeds.

### Screenshots

N/A — no UI changes.

### Related issue(s)

No related issue — small follow-up to #672.

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/web/scripts/redeemDelay.ts
+++ b/web/scripts/redeemDelay.ts
@@ -13,13 +13,14 @@ import {
   readContract,
   setBalance,
   stopImpersonatingAccount,
-  waitForTransactionReceipt,
   writeContract,
 } from "viem/actions";
 import { mainnet } from "viem/chains";
 
 import { gatewayAbi } from "../../packages/gateway/src/abi/gatewayAbi.ts";
 import { gatewayAddresses } from "../../packages/gateway/src/gatewayAddresses.ts";
+
+import { confirmTransaction } from "./utils.ts";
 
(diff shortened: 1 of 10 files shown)


gabmontes (commented):
LGTM but the e2e tests started to fail.

gndelia:
> LGTM but the e2e tests started to fail.

Then that's good - the test was a bit flaky then haha

gndelia:
@gabmontes  tests fixed in [e0cfe37](https://github.com/vetro-protocol/vetro-monorepo/pull/676/commits/e0cfe378b142cd093a34067019d6b7da42a67657)

gabmontes (approved):

ArturDolzan (approved):

Issue #445, opened 2026-06-03
Create CLI for agents to operate on Vetro


gndelia:
### Feature or change description

Create a CLI that sits on top of our packages to allow agents to operate within Vetro context

gndelia:
Things to keep in mind

- [ ] Allow users to save some config for custom rpcs: For example: something like `vetro-cli config rpcUrl <some-url>` and get's saved locally. This will be useful as the default RPC from `viem` for Ethereum does not work :Agitate:
- [x] Check for swaps to be enabled / paused (depends on https://github.com/vetro-protocol/vetro-monorepo/issues/645 )
- [ ] Add `publishConfig` and `repository` so the CLI can be published with `provenance`

PR #679, opened 2026-07-31
Add CLI package info

gndelia:
### Description

This PR adds information needed for enabling `provenance` once the package is released through CI. Note that `provenance: true` is still missing on `publishConfig` - it will be added once the package is stable. I also removed `private: true` bc the package is already published in beta so not needed anymore

### Screenshots

N/A

### Related issue(s)

Related to #445

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -11,6 +11,11 @@
   "bin": {
     "vetro-cli": "./_esm/cli.js"
   },
+  "repository": {
+    "directory": "packages/cli",
+    "type": "git",
+    "url": "git+https://github.com/vetro-protocol/vetro-monorepo.git"
+  },
   "scripts": {
     "build": "tsc --noEmit",
     "bundle": "node esbuild.ts",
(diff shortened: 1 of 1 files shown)


packages/cli/package.json
```diff
@@ -41,6 +46,8 @@
   "engines": {
     "node": ">=24"
   },
-  "private": true,
```
gabmontes, line 44: For consistency with the other published packages (bridge, earn, gateway, treasury), which all keep "private": false explicitly, consider setting "private": false here instead of removing the field. Same effect, but keeps the publish intent explicit across the monorepo.

— Gabi 🤖

gndelia: Added in [0a97f9f](https://github.com/vetro-protocol/vetro-monorepo/pull/679/commits/0a97f9f52a6e0ca0182ddd59b0aa2412fd7bff78)

gabmontes (approved):

ArturDolzan (approved):

Issue #659, opened 2026-07-28
Consider a "core" package


gndelia:
Now that CLI is comnig (See #445), I find that several small pieces are duplicated accross web, api, scripts folder and now cli too (either through all, or some)

I'm listing here these options as I find them - in case there are enough, it may be worth creating a "core" / "sdk" package to hold these

- [ ] Slippage functionality (amounts allowed, default slippage)
- [x] Token list
- [ ] Mapping a gateway to a staking vault (they're not solved through contracts)
- [ ] Mapping a whitelisted token <-> pegged token (it's a combination of a few contract calls)
- [ ] Common types across the whole monorepo (`Token`, common `EvenType` from packages)
- [ ] Some utilities, such as `isAddressValid`, `updateRpcUrls`, `createClient` with fallback/cache/batching

PR #680, opened 2026-07-31
Fail swap mint when deposits are paused

gndelia:
### Description

The CLI now checks if the minting is paused, as the web did so in #955
Note that part of the helpers are hope to be removed by #659 or similar as yes - some code is repeated

### Screenshots

N/A

### Related issue(s)

Related to #445

### Checklist

- [ ] Manual testing passed.


packages/cli/src/features/swap/commands/mint.ts
```diff
@@ -67,15 +69,31 @@ export function register(swap: Command) {
 
       const amountIn = parseUnits(options.amount, tokenIn.decimals);
 
-      const [peggedTokenOut, maxMint] = await Promise.all([
+      const isDepositActive = async function () {
+        const treasury = await getTreasury(client, {
+          address: tokenIn.gatewayAddress,
+        });
+        const [, , , depositActive] = await getTokenConfig(client, {
```
gabmontes, line 76: **Optional nit:** the sparse positional destructuring `const [, , , depositActive]` is a bit fragile — if the `tokenConfig` struct field order ever changes, this silently reads the wrong field with no type error. The web hook (`useTokenConfig`) names all six positions, which reads better and keeps both call sites consistent. Consider naming the leading fields too, e.g. `const [vault, oracle, stalePeriod, depositActive] = ...`.

— Gabi 🤖

gndelia: Naming the leading fields wouldn't change the problem of struct changing. But that will require a contract upgrade (And a breaking change), so I don't see that happening any time soon.

gabmontes (approved):

ArturDolzan (approved):

PR #682, opened 2026-08-03
Replace isValidUrl with native URL.canParse

gndelia:
### Description

Came across `URL.canParse()` and realized it does exactly what our custom `isValidUrl` helper in `web/src/utils/url.ts` did — a `try { new URL(x) } catch` wrapper — so there's no reason to keep our own version. This drops the helper and calls the native static method at the 16 call sites instead.

The `isValidUrl` unit tests were removed along with it; they asserted platform behavior rather than our own code.

One note: `URL.canParse` does raise the minimum browser we support to Chrome 120 / Firefox 115 / Safari 17, a bit above Vite's current default baseline. That said, all three shipped in 2023 and are ~2.5–3 years old by now, and the method is available in Node 18.17+ / 20+ as well, so we're fine relying on it.

### Screenshots

N/A — no UI changes.

### Related issue(s)

N/A

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/web/playwright.config.ts
+++ b/web/playwright.config.ts
@@ -30,7 +30,7 @@ export default defineConfig({
     env: {
       // Swap/redeem is fully on-chain; the portal API only supplies USD display
       // values. Disable it so e2e stays hermetic (prices render as $0, which
-      // these tests don't assert on). isValidUrl("") is false → query disabled.
+      // these tests don't assert on). URL.canParse("") is false → query disabled.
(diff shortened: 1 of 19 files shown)


gndelia:
All comments from Copilot are exactly the same - we'll see from Sentry if `URL.canParse` not being defined is a problem, but all in all we should be greatly covered.

gabmontes (approved):

ArturDolzan (approved):

PR #683, opened 2026-08-03
Add --rpc-url and stamp calldata with the endpoint's chain

gndelia:
### Description

Adds a global `--rpc-url <url>` option, so a custom RPC endpoint can be set per invocation instead of only through the `RPC_URL` env var. It is accepted anywhere on the command line and resolves `--rpc-url` → `RPC_URL` → viem's public mainnet RPC.

Adding it surfaced a bug: the emitted `chainId` was hardcoded to mainnet while reads came from whatever `RPC_URL` pointed at, so calldata built against a fork was still labelled `"0x1"`. It is now read from the endpoint with `eth_chainId`.

Vetro is mainnet-only for now, so that chain id is also validated before any contract read — anything other than mainnet (`1`) or a local fork (`31337`) fails fast instead of returning state from the wrong chain.

Also fixes a leak the flag exposed: commander echoes a rejected option value, so an API key in `RPC_URL` reached stderr. Usage errors are now redacted too.

**Commits:**

5a66f91 Read chainId from the RPC, don't assume mainnet
4ad83b5 Add a global --rpc-url and guard the chain

### Screenshots

N/A — CLI only, no UI changes.

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/cli/src/lib/args.ts
+++ b/packages/cli/src/lib/args.ts
@@ -30,6 +30,19 @@ export function parseGateway(value: string) {
   return address;
 }
 
+export function parseRpcUrl(value: string) {
+  if (!URL.canParse(value)) {
+    throw new InvalidArgumentError(`Invalid RPC URL: "${value}"`);
+  }
+  const { protocol } = new URL(value);
+  if (protocol !== "http:" && protocol !== "https:") {
+    throw new InvalidArgumentError(
+      `RPC URL must use http or https: "${value}"`,
+    );
+  }
+  return value;
+}
+
 export function parseSlippage(value: string) {
   const normalized = value.replace(",", ".");
   if (normalized !== "" && !/^\d+(\.\d?)?$/.test(normalized)) {
--- a/packages/cli/src/lib/args.ts
+++ b/packages/cli/src/lib/args.ts
@@ -31,11 +31,13 @@ export function parseGateway(value: string) {
 }
 
 export function parseRpcUrl(value: string) {
-  if (!URL.canParse(value)) {
+  let url: URL;
+  try {
+    url = new URL(value);
+  } catch {
     throw new InvalidArgumentError(`Invalid RPC URL: "${value}"`);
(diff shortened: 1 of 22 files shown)


ArturDolzan (dismissed):
uACK

packages/cli/src/lib/args.ts
```diff
@@ -30,6 +30,19 @@ export function parseGateway(value: string) {
...
+    throw new InvalidArgumentError(`Invalid RPC URL: "${value}"`);
+  }
+  const { protocol } = new URL(value);
+  if (protocol !== "http:" && protocol !== "https:") {
+    throw new InvalidArgumentError(
+      `RPC URL must use http or https: "${value}"`,
+    );
+  }
+  return value;
+}
```
gabmontes, line 46: `parseRpcUrl` parses the URL twice — `URL.canParse(value)` and then `new URL(value)`. A single parse in a `try/catch` avoids the redundant work and keeps the validity + protocol checks together:

```ts
export function parseRpcUrl(value: string) {
  let url: URL;
  try {
    url = new URL(value);
  } catch {
    throw new InvalidArgumentError(`Invalid RPC URL: "${value}"`);
  }
  if (url.protocol !== "http:" && url.protocol !== "https:") {
    throw new InvalidArgumentError(`RPC URL must use http or https: "${value}"`);
  }
  return value;
}
```

Minor / non-blocking.

— Auto Gabi

gndelia: Fixed in [6f7c9cc](https://github.com/vetro-protocol/vetro-monorepo/pull/683/commits/6f7c9cc6ed202d6ce2430903006fc20a4cc29f62). `parseRpcUrl` now parses once in a `try`/`catch`, matching `parseAddress` right above it.

ArturDolzan (approved):

gabmontes (approved):

PR #684, opened 2026-08-03
Track Uniswap V3 pools in the DEX tab

### LINKED ISSUE #665: Add Uniswap pools to the DEX tab

These pools have been created on Uniswap

WBTC/VUSD v3 pool
https://app.uniswap.org/explore/pools/ethereum/0x8B776DEA1efd17fE1c6B49603F9093f856aD8073

vetBTC/WBTC
https://app.uniswap.org/explore/pools/ethereum/0x3467593228DBd4AcE9E44D58A193Cb5DC7A3efEc

vetBTC/cbBTC
https://app.uniswap.org/explore/pools/ethereum/0x3656E05FcaCad70f2f12893CF7c215578B6f8580

hemiBTC/vetBTC
https://app.uniswap.org/explore/pools/ethereum/0x1335D0AeAE12eDf259B0Ffb9B475d0df99d0E83c

We should list them in the dashboard

gndelia:
### Description

Adds Uniswap as a third DEX source in the internal dashboard, alongside Curve and Sushi, so the pools listed in #665 (WBTC/VUSD, vetBTC/WBTC, vetBTC/cbBTC, hemiBTC/vetBTC) show up in the DEX tab.

Pools are discovered on-chain rather than hardcoded: tracked tokens are paired against the collateral each gateway whitelists (gateway → treasury → `getWhitelistedTokens`) and looked up in the V3 factory across the four mainnet fee tiers. Per-pool supplies, token prices and 24h volume come from Uniswap's GraphQL API, reached through a new `/api/uniswap` worker proxy — the existing Sushi proxy generalized, since both upstreams reject the deployed browser's cross-origin requests.

Two figures are derived rather than taken as published: TVL, because Uniswap counts only the legs it can price, and 24h fees, which it doesn't expose at all and follow from volume × fee tier.

The Stats distribution cards now also carry a venue badge per pool, in the legend and in the pie tooltip, so it's clear which protocol each slice of a token's liquidity sits on.

### Screenshots

https://github.com/user-attachments/assets/760904e7-2737-471b-a51d-5b5c5f4711da

### Related issue(s)

Closes #665

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/internal-dashboard/src/lib/uniswapApi.ts
+++ b/internal-dashboard/src/lib/uniswapApi.ts
@@ -94,11 +94,16 @@ export const fetchUniswapPoolData = async function (
     headers: { "content-type": "application/json" },
     method: "POST",
   });
+  // GraphQL reports failures as a 200 with an `errors` array, which can arrive
+  // alongside a partially populated `data`; surface the real message rather than
+  // silently pricing the pool off whatever fields did come back.
+  if (body.errors?.length) {
+    const reason = body.errors.map((error) => error.message).join("; ");
+    throw new Error(`Uniswap API error for pool ${address}: ${reason}`);
+  }
   const pool = body.data?.v3Pool;
   if (!pool) {
-    const reason =
-      body.errors?.map((error) => error.message).join("; ") ?? "not found";
-    throw new Error(`Uniswap pool ${address}: ${reason}`);
+    throw new Error(`Uniswap pool ${address} not found`);
   }
   return {
     feeTier: pool.feeTier,
--- a/internal-dashboard/src/lib/uniswapApi.ts
+++ b/internal-dashboard/src/lib/uniswapApi.ts
@@ -1,0 +1,112 @@
+import fetch from "fetch-plus-plus";
+import { type Address } from "viem";
+
+const UNISWAP_API_URL = "/api/uniswap";
+
+const CHAIN = "ETHEREUM";
+
+const query = `
+  query V3Pool($address: String!, $chain: Chain!) {
+    v3Pool(address: $address, chain: $chain) {
+      feeTier
+      priceHistory(duration: WEEK) {
+        token1Price
+      }
+      token0 {
+        market(currency: USD) {
+          price {
+            value
+          }
+        }
+      }
+      token0Supply
+      token1 {
+        market(currency: USD) {
+          price {
+            value
+          }
+        }
+      }
+      token1Supply
+      volume24h: cumulativeVolume(duration: DAY) {
+        value
+      }
+    }
+  }
+`;
+
+type RawToken = { market: { price: { value: number } | null } | null };
+
+type RawV3Pool = {
+  feeTier: number;
+  priceHistory: { token1Price: number | null }[] | null;
+  token0: RawToken;
+  token0Supply: number | null;
+  token1: RawToken;
+  token1Supply: number | null;
+  volume24h: { value: number } | null;
+};
+
+type UniswapPoolData = {
+  feeTier: number; // hundredths of a bip (3000 = 0.3%)
+  supplies: [number, number]; // whole tokens held
+  usdPrices: [number, number];
+  volumeUsd24h: number;
+};
+
+const poolRate = function (history: RawV3Pool["priceHistory"]) {
+  const rates = (history ?? [])
+    .map((point) => point.token1Price ?? 0)
+    .filter((rate) => rate > 0);
+  return rates[rates.length - 1] ?? 0;
+};
+
+const resolvePrices = function ({
+  prices,
+  rate,
+}: {
+  prices: [number, number];
+  rate: number;
+}): [number, number] {
+  const [price0, price1] = prices;
+  if (rate <= 0) {
+    return prices;
+  }
+  if (price0 > 0 && price1 <= 0) {
+    return [price0, price0 / rate];
+  }
+  if (price1 > 0 && price0 <= 0) {
+    return [price1 * rate, price1];
+  }
+  return prices;
+};
+
+const tokenPrice = (token: RawToken) => token.market?.price?.value ?? 0;
+
+export const fetchUniswapPoolData = async function (
+  address: Address,
+): Promise<UniswapPoolData> {
+  const body: {
+    data?: { v3Pool: RawV3Pool | null };
+    errors?: { message: string }[];
+  } = await fetch(UNISWAP_API_URL, {
+    body: JSON.stringify({ query, variables: { address, chain: CHAIN } }),
+    headers: { "content-type": "application/json" },
+    method: "POST",
+  });
+  const pool = body.data?.v3Pool;
+  if (!pool) {
+    const reason =
+      body.errors?.map((error) => error.message).join("; ") ?? "not found";
+    throw new Error(`Uniswap pool ${address}: ${reason}`);
+  }
+  return {
+    feeTier: pool.feeTier,
+    supplies: [pool.token0Supply ?? 0, pool.token1Supply ?? 0],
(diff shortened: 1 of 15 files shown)


internal-dashboard/src/lib/uniswapApi.ts
```diff
@@ -0,0 +1,112 @@
...
+  supplies: [number, number]; // whole tokens held
+  usdPrices: [number, number];
+  volumeUsd24h: number;
+};
+
+const poolRate = function (history: RawV3Pool["priceHistory"]) {
+  const rates = (history ?? [])
+    .map((point) => point.token1Price ?? 0)
+    .filter((rate) => rate > 0);
+  return rates[rates.length - 1] ?? 0;
```
gabmontes, line 61: `priceHistory(duration: WEEK)` is read as ascending here — `rates[rates.length - 1]` assumes the most recent point is last. If the gateway returns points newest-first, the fallback cross-price would be up to a week stale. Worth confirming the ordering, and sorting by timestamp to be safe.

— Auto Gabi

internal-dashboard/src/lib/uniswapApi.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (rate <= 0) {
+    return prices;
+  }
+  if (price0 > 0 && price1 <= 0) {
+    return [price0, price0 / rate];
+  }
+  if (price1 > 0 && price0 <= 0) {
+    return [price1 * rate, price1];
+  }
+  return prices;
```
gabmontes, line 81: When neither leg has a Uniswap USD market price this returns `[0, 0]`, so `tvlUsd` (and the fee-derived `baseApy`) collapse to 0 and the pool renders unpriced. That hits a tracked/tracked pair like hemiBTC/vetBTC from #665, while WBTC/cbBTC-anchored pairs are fine. Since the dashboard already has its own oracle (`useTokenPrices` / Portal), should this fall back to it instead of showing $0?

— Gabi 🤖

internal-dashboard/src/lib/uniswapApi.ts
```diff
@@ -0,0 +1,112 @@
...
+    body: JSON.stringify({ query, variables: { address, chain: CHAIN } }),
+    headers: { "content-type": "application/json" },
+    method: "POST",
+  });
+  const pool = body.data?.v3Pool;
+  if (!pool) {
+    const reason =
+      body.errors?.map((error) => error.message).join("; ") ?? "not found";
+    throw new Error(`Uniswap pool ${address}: ${reason}`);
+  }
```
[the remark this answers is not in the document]
gndelia: Fixed in [7586cdb](https://github.com/vetro-protocol/vetro-monorepo/pull/684/commits/7586cdb). `body.errors` is now checked first and throws with the real message, leaving a dedicated "not found" for a null `v3Pool` — the same two-step shape `sushiApi.ts` uses.

internal-dashboard/src/lib/uniswapV3.ts
```diff
@@ -0,0 +1,78 @@
...
+  return [...pairs.values()].flatMap((pair) =>
+    uniswapV3FeeTiers.map((fee) => ({ ...pair, fee })),
+  );
+};
+
+export const findPools = async function (queries: PoolQuery[]) {
+  const addresses = await multicall(client, {
+    allowFailure: false,
+    batchSize: 0,
+    contracts: queries.map((query) => ({
```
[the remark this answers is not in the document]
gndelia: `batchSize: 0` is viem's documented way to disable calldata chunking, not an invalid value. In viem 2.44.4 the chunk split is guarded by `batchSize > 0 && currentChunkSize > batchSize`, so `0` simply means "one `aggregate3` for all calls" — there is no error path on it. It is also honored here: the client-level override at `multicall.js:54` only applies when `client.batch.multicall` is an *object*, and `client.ts` sets it to `true`.

Keeping it, because the default measurably regresses this call. Discovery is currently 120 `getPool` calls (10 candidate tokens → 30 tracked-involving pairs × 4 fee tiers), and the 1024-byte default splits that into about a dozen chunks that viem fires concurrently via `Promise.allSettled`. Against the configured RPC (`eth.drpc.org`) that burst gets rate-limited:

```
batchSize 0   -> OK 3286 ms | OK 403 ms
default 1024  -> FAILED: HTTP request failed. | OK 2627 ms
```

Since `allowFailure: false` is set, one rejected chunk throws away the entire discovery and the Uniswap source disappears from the dashboard rather than degrading.

That said, the confusion is fair — a bare `0` explains none of this. Added a comment in [335d4fb](https://github.com/vetro-protocol/vetro-monorepo/pull/684/commits/335d4fb).

internal-dashboard/src/lib/uniswapApi.ts
gndelia: Confirmed ascending. Queried the gateway the worker proxies to (`interface.gateway.uniswap.org`) for USDC/WETH 0.05%:

```
count: 168, first: 1785186000, last: 1785787200, ascending: true, now: 1785789415
```

168 hourly points across the week, strictly ascending by `timestamp`, with the last point ~37 minutes old — so `rates[rates.length - 1]` is the most recent, not week-stale.

Leaving the sort out: `timestamp` is not in the selection set today, so adding one would mean widening the query to guard against a hypothetical reordering by a third party. And this value is only a fallback cross-rate, used when one leg has no Uniswap USD price.

internal-dashboard/src/lib/uniswapApi.ts
gndelia: Confirmed the symptom. Of the four pools discovery finds today, hemiBTC/vetBTC (`0x1335D0AeAE12eDf259B0Ffb9B475d0df99d0E83c`) is the one where Uniswap prices neither leg, so it renders at $0 TVL / 0% APY:

| Pool | token0 | token1 |
| --- | --- | --- |
| WBTC/VUSD 0.3% | $63,848 | $0 |
| WBTC/vetBTC 0.05% | $63,848 | $0 |
| cbBTC/vetBTC 0.05% | $63,815 | $0 |
| hemiBTC/vetBTC 0.05% | $0 | $0 |

(Small note: hemiBTC is whitelisted collateral rather than a tracked token, but that does not change the outcome.)

Keeping the $0 rather than falling back to Portal. The Uniswap entries should reflect what Uniswap itself reports about the pool — substituting our own oracle would make two pools with identical Uniswap data render differently depending on whether Portal happens to cover a leg, and $0 is a truthful signal that Uniswap has no USD market for that pair rather than a display bug. If that reads as confusing in the UI, the better fix is an explicit "unpriced" state, not a different oracle's number.

internal-dashboard/src/lib/uniswapApi.ts
```diff
@@ -0,0 +1,117 @@
...
+  supplies: [number, number]; // whole tokens held
+  usdPrices: [number, number];
+  volumeUsd24h: number;
+};
+
+const poolRate = function (history: RawV3Pool["priceHistory"]) {
+  const rates = (history ?? [])
+    .map((point) => point.token1Price ?? 0)
+    .filter((rate) => rate > 0);
+  return rates[rates.length - 1] ?? 0;
```
ArturDolzan, line 61: Wonder if `rates.at(-1) ?? 0` reads a touch cleaner than the manual `length - 1` index here — same behavior for the empty case. Total nit, feel free to ignore.


Artur's 🤖

gndelia: I've been using `[.length - 1]` all my life so I think it's as readable as the other option 😅

gabmontes (approved):
uACK

ArturDolzan (approved):

Issue #646, opened 2026-07-24
Target Yield Earn (VUSDx)


gndelia:
### Feature or change description

Umbrella issue for `@vetro-protocol/target-yield-earn` in which I'll be sending changes to start with this feature

- [x] Create `viem-erc7540` package for async erc-4626 vaults support https://github.com/hemilabs/viem-erc7540/pull/1
- [ ] Update `docs/Domain`

Create packge for VUSDx. It should use `viem-erc7540` package, as well as the following list of read-only methods:
- [x] `epochId()`
- [ ] getEpoch(uint256 epochId_)
- [x]  `currentRate()`  (only for the existing current rate while epoch is live, otherwise, we'll need `rate(uint256 epochId_)`)
- [x] `maxRequestDeposit(address controller_)` 
- [x]  `maxRequestRedeem(address owner_)`
- [ ] `cancelDepositRequest(controller)`
- [ ]  `paused` (This actually comes from [PausableUpgradeable.sol extension from OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/184d9fe75bcd640ad8e0bc86dcef1858fb33cd13/contracts/utils/PausableUpgradeable.sol#L85)
- [ ] `isShutdown()`
- [ ] `isTerminated()`

## Notes

The following list of values are hardcoded in the contract (therefore we can update them in our code with `initialData` + `revalidate`)

```solidity
uint64 public constant MIN_EPOCH_DURATION = 15 days;
uint64 public constant MIN_EXIT_WINDOW = 1 days;
uint64 public constant MAX_EXIT_WINDOW = 7 days;
```

PR #691, opened 2026-08-05
Add maxRequestRedeem read to target-yield-earn

gndelia:
### Description

Adds `getMaxRequestRedeem` to `@vetro-protocol/target-yield-earn`, wrapping the epoch vault's `maxRequestRedeem(owner)` view.

It returns the shares an owner can still request for redemption — their share balance — and `0` whenever a `requestRedeem` call would revert: outside the exit window, while shutdown, or when a redeem batch from an earlier epoch is still pending and the vault hasn't terminated. Unlike the deposit read, it does not gate on `paused()`, and termination does not zero it.

The parameter is `owner` rather than the deposit read's `controller`, matching the contract, the spec, and `maxRedeem` in ERC-4626.

> [!NOTE]
> Stacked on #690 — based on `add-max-request-deposit`, so review that one first. The diff here is the second commit only.

### Screenshots

N/A — no UI changes.

### Related issue(s)

Related to #646

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/packages/target-yield-earn/README.md
+++ b/packages/target-yield-earn/README.md
@@ -22,6 +22,7 @@ import {
   getCurrentRate,
   getEpochId,
   getMaxRequestDeposit,
+  getMaxRequestRedeem,
   pendingDepositRequest,
 } from "@vetro-protocol/target-yield-earn/actions";
 import { createPublicClient, http } from "viem";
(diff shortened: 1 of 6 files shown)


ArturDolzan (dismissed):

gabmontes (dismissed):

gndelia:
Rebased now that the stack was updated

ArturDolzan (approved):

gabmontes (approved):

Issue #695, opened 2026-08-05
Review and fix how we import files for the different formats


gndelia:
TypeScript, JavaScript, and Node have different flavors for importing files when using ESM. This mix can cause problems, as sometimes don't work properly. Additionally, Vite, Playwright, and Vitest  solve this partially internally in their own implementations, so that's why this issue does not come up until we include a raw `node script.ts` into the mix

Assuming the current struture of this repo, I think we should support all these cases:

1. Run a script from Node, like "node ./web/scripts/foo.ts"
2. These scripts can be called from E2E tests (say `swap.test.ts`) to set up the scenario that needs to be tested
3. These scripts can be called by AI agents when using Playwright to run the app for whatever reason
4. These scripts can import any file from the packages, ideally, using the "exports" from the package.json. So for example, from `foo.ts` we should be able to do `import { someting } from '@vetro-protocol/bridge'`. For the packages that follow the `viem` format, we should follow the `actions` mapping (so `import { someAction } from '@vetro-protocol/bridge/actions'` should work as well)
5. `api` and `web` (And, technically, any other workspace) can use the packages under `packages/*`. For this, they will use the `workspace:*` option from `pnpm`
6. Any workspace under `packages/*` can be built into JavaScript, published into NPM, and have scripts in TypeScript (outside of this repo) to install them and use them as in step 4.
7. The CLI can use  the packages, and have the CLI invoked from the terminal (either by devs or by agents).

I think all these scenarios need to be solved before implementing #677 and #659 and to improve how #445 is implemented too

PR #698, opened 2026-08-07
Bundle earn and take isAddressValid from core

gndelia:
### Description

This PR updates the `earn` package, so it uses the newly created '`core` package in #697. For that, a small `esbuild` script was added that takes care of inlining `core` as it won't be published, and TypeScript alone isn't enough to inline references from `workspace:*` from `pnpm`

The addition of using `.ts` extensions instead of `.js` imports makes the imports runnable from the terminal when using `node foo.ts` -  this makes it so that the scripts can use the packages. For TypeScript to allow that, we need `allowImportingTsExtensions: true` in `tsconfig.json`.

This makes it all work. For consumers of the package from NPM, it all works the same as before. Now, we can also run scripts which use the packages which internally have `.ts` imports. Running these files from Playwright, E2E tests, the api or the web continues to work, thus covering all scenarios mentioned in #695

`isAddressValid.ts` is now read from `core` so it can be dropped

### Related issue(s)

Related to #696 #659

### Checklist

- [ ] Manual testing passed.


scripts/bundle-package.js
```diff
@@ -0,0 +1,52 @@
...
+  isPrivateWorkspacePackage,
+);
+if (misplaced.length > 0) {
+  throw new Error(
+    `${misplaced.join(", ")} is private and gets inlined, so it must be a devDependency -- listing it under "dependencies" publishes a package npm cannot install.`,
+  );
+}
+
+// Private workspace need to be inlined - so mark the rest as "external"
+// so they are included as part of the source code.
```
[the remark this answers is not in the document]
gndelia: Fixed in [e875761c](https://github.com/vetro-protocol/vetro-monorepo/pull/698/commits/e875761c06f39f0dca2ecf9ab024223af691640b). The comment now says private workspace packages are the ones inlined, and everything in `external` stays a runtime import.

scripts/bundle-package.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Private workspace packages are not published, so they must be inlined.
+// Everything else stays "external" - kept as a runtime import, not bundled.
+const external = Object.keys({
+  ...packageJson.dependencies,
+  ...packageJson.peerDependencies,
+})
+  .filter((name) => !isPrivateWorkspacePackage(name))
+  .flatMap((name) => [name, `${name}/*`]);
```
gabmontes: The wallet actions import `EventEmitter` from the `events` node builtin, but `external` is derived only from `dependencies`/`peerDependencies`. Under `platform: "neutral"`, esbuild can't resolve `events` and `pnpm bundle` fails (reproduced locally: `Could not resolve "events"`). Since `bundle` only runs in `prepublishOnly`, CI stays green but publishing breaks. Please add node builtins to `external` (e.g. `builtinModules` from `node:module`, including their `node:` forms).

Gabi 🤖

packages/earn/package.json
```diff
@@ -17,11 +17,12 @@
   },
   "scripts": {
     "build": "tsc --noEmit",
-    "build:esm": "tsc -p tsconfig.build.json --outDir ./_esm --sourceMap",
     "build:types": "tsc -p tsconfig.build.json --declarationDir ./_types --emitDeclarationOnly --declaration --declarationMap",
+    "prebundle": "pnpm build",
+    "bundle": "node ../../scripts/bundle-package.js",
     "clean": "rm -rf ./_esm ./_types",
     "prepack": "cp ../../LICENSE LICENSE",
```
gabmontes, line 24: `bundle` and `build:types` only run in `prepublishOnly`, so CI never exercises the published output. This hides both the `events` bundling failure noted in `scripts/bundle-package.js` and any issues in the `.d.ts` emitted under `allowImportingTsExtensions`. Consider running the bundle in CI to catch publish-time breakage.

Auto Gabi

scripts/bundle-package.js
gndelia: image


`pnpm run bundle` doesn't fail locally 🤔

packages/earn/package.json
gndelia: I updated the scripts so bundle is run after build [6047d86](https://github.com/vetro-protocol/vetro-monorepo/pull/698/commits/6047d861ec22c89f14e821351f6705d348203d57)

packages/earn/src/stakingVaultAddresses.ts
```diff
@@ -1,9 +1,15 @@
+import {
+  sVetBtcAddress as coreSVetBtcAddress,
+  sVusdAddress as coreSVusdAddress,
+  stakingVaultAddresses as coreStakingVaultAddresses,
+} from "@vetro-protocol/core";
```
ArturDolzan, line 5: Perhaps it's worth peeking at the generated `_esm` chunk — going through core's barrel seems to drag `knownTokens` (~7 KB, unreferenced) and a `viem/chains` import into every earn entry point, since esbuild can't prove `chainId: mainnet.id` is pure. Splitting the addresses into their own core module took the shared chunk from 7,339 B down to 563 B in a quick local test.


Artur's 🤖

scripts/bundle-package.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Private workspaces are only allowed in devDependencies.
+const misplaced = Object.keys(packageJson.dependencies ?? {}).filter(
+  isPrivateWorkspacePackage,
+);
+if (misplaced.length > 0) {
+  throw new Error(
+    `${misplaced.join(", ")} is private and gets inlined, so it must be a devDependency -- listing it under "dependencies" publishes a package npm cannot install.`,
+  );
+}
```
ArturDolzan: Not sure it's worth it, but since `external` already spreads `peerDependencies`, could be worth having the private-workspace guard look there too — otherwise that's the one field where a private package slips through and gets inlined while still being declared.


Artur's 🤖

packages/earn/src/stakingVaultAddresses.ts
gndelia: Fixed in [08c007b4](https://github.com/vetro-protocol/vetro-monorepo/pull/698/commits/08c007b4). Confirmed with a clean bundle both ways: the shared chunk goes from 7,339 B (carrying `knownTokens` and a `viem/chains` import) down to 575 B. Core's `sideEffects: false` wasn't enough on its own, since esbuild can't prove the `knownTokens` literal is pure. The addresses now live in `packages/core/src/stakingVaultAddresses.ts`.

scripts/bundle-package.js
gndelia: Fixed in [bf634f4d](https://github.com/vetro-protocol/vetro-monorepo/pull/698/commits/bf634f4d). The guard now spreads `peerDependencies` alongside `dependencies`, matching what `external` reads. Verified by temporarily declaring core as a peer of earn: it throws now, and silently passed before. No package hits this today, so it's closing the hole rather than fixing a live break.

scripts/bundle-package.js
```diff
@@ -0,0 +1,53 @@
...
+
+// Private workspace packages are not published, so they must be inlined.
+// Everything else stays "external" - kept as a runtime import, not bundled.
+const external = Object.keys({
+  ...packageJson.dependencies,
+  ...packageJson.peerDependencies,
+})
+  .filter((name) => !isPrivateWorkspacePackage(name))
+  .flatMap((name) => [name, `${name}/*`]);
```
ArturDolzan: Now that the guard also spreads `peerDependencies` (`bf634f4`), wonder if the `!isPrivateWorkspacePackage` filter on `external` can still drop anything — looks like the throw above already rules that out. Might be worth hoisting the shared `Object.keys` too?


Artur's 🤖

gndelia: Fixed in [8ddaaca2](https://github.com/vetro-protocol/vetro-monorepo/pull/698/commits/8ddaaca2). Right on both counts — the throw rules out every private name before `external` is built, and `isPrivateWorkspacePackage` returns a strict boolean, so the two filters were exact complements with no truthiness gap. Hoisted the key list into `declaredDependencies` and dropped the filter; the comment above `external` described the deleted code, so it's been rewritten. Verified the earn bundle is byte-identical afterwards, and the guard still throws when core is declared as a peer.

ArturDolzan (approved):
LGTM!

gabmontes (approved):
LGTM

PR #706, opened 2026-08-10
Fix TVL for Sushi price-band views

### LINKED ISSUE #704: TVL in range in pools from Dex is incorrectly calculated
### Description

The sushi pool shows for the total range the following TVL

Image

For the range of `0.96-1.04`, which is included in the above pool, it's showing a large TVL, which is impossible

Image

gndelia:
### Description

The `$0.96–$1.04` view on the Sushi VUSD/USDT pool reported a TVL larger than the pool's own full range — ~$1.8M inside a ~$43k pool.

`computeBandAmounts` took the pool's current liquidity (the `L` active at the current tick) and spread it uniformly across the whole band. That `L` belongs to positions clustered around the price — roughly ±$0.0008 for this pool — so extrapolating it across a ±4% band overstated the amounts by ~43x. The value was unbounded relative to the pool's real reserves.

The band is now cut at every initialized tick it contains, and each slice valued with the liquidity actually active across it:

- **`src/lib/v3PoolState.ts`** (new) — reads `slot0`, `liquidity`, `tickSpacing`, the `tickBitmap` words spanning the band, and `ticks()` for each initialized tick found. Three multicalls; two words and 15 ticks for the configured band. No data API publishes the per-tick distribution (Sushi's GraphQL schema has no tick fields), so it has to come from chain.
- **`src/lib/v3PositionMath.ts`** — `computeBandAmounts` walks down from the pool's reported liquidity to recover the liquidity in the band's lowest slice, then walks up applying each tick's `liquidityNet`. Adds `priceToTick` and the `TickLiquidity` type.
- **`src/fetchers/fetchSushiPools.ts`** — one read spanning all configured bands so they share a snapshot; pools without bands skip it entirely.
- **`src/lib/sushiApi.ts`** — drops the now-unused `liquidity` / `sqrtPrice` fields.

Verified against the pool's own [Sushi page](https://www.sushi.com/ethereum/pool/v3/0x6c2bd2f9711f204e595d334c6b7b672851b7d699):

- Walking every tick reproduces their reported 25,861 VUSD / 27,305 USDT to within 0.03%, with `netSum` over all 16 initialized ticks landing at exactly 0.
- The outermost ticks read convert to 0.833945 and 1.2001 — exactly the bounds of their depth chart.
- A narrow read window (0.96–1.04) and a wide one (0.2–5) give byte-identical band amounts at the same block. Locked in as a unit test, since it's what makes the cheap windowed read safe.

The band now reads ~$38.8k of the pool's ~$53.2k, i.e. ~73% of liquidity within ±4% of peg.

### Screenshots

image

### Related issue(s)

Closes #704

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/internal-dashboard/src/fetchers/fetchSushiPools.ts
+++ b/internal-dashboard/src/fetchers/fetchSushiPools.ts
@@ -3,7 +3,8 @@ import { formatUnits } from "viem";
 import { type SushiPoolConfig, sushiPools } from "../config/sushiPools";
 import { fetchSushiPoolData, type SushiToken } from "../lib/sushiApi";
 import { type PoolCoin, type TrackedPool } from "../lib/types";
-import { computeBandAmounts } from "../lib/v3PositionMath";
+import { fetchV3PoolState } from "../lib/v3PoolState";
+import { computeBandAmounts, priceToTick } from "../lib/v3PositionMath";
 
 const buildCoins = ({
   balances,
@@ -23,13 +24,14 @@ const buildCoins = ({
   }));
 
 // The Sushi implementation of a DEX pool source. One GraphQL call to Sushi's own
-// API (fetchSushiPoolData) supplies everything — token identity, balances, TVL,
-// prices, 24h volume / fees / APR, plus the pool's current liquidity / sqrt price
-// — so nothing is read on-chain. A configured price band's concentrated liquidity
-// is derived from that current liquidity with Uniswap's own Position math
-// (lib/v3PositionMath). USD prices anchor the reference leg (the non-tracked
-// stable) at $1 and take the tracked leg's price from the pool rate. Gauge
-// emissions aren't a Sushi concept, so that stays unset.
+// API (fetchSushiPoolData) supplies everything the whole-pool entry needs — token
+// identity, balances, TVL, prices, 24h volume / fees / APR. A configured price
+// band additionally needs the pool's liquidity spread across ticks, which no data
+// API publishes, so that much is read on-chain (lib/v3PoolState) and turned into
+// token amounts with Uniswap's own Position math (lib/v3PositionMath). USD prices
+// anchor the reference leg (the non-tracked stable) at $1 and take the tracked
+// leg's price from the pool rate. Gauge emissions aren't a Sushi concept, so that
+// stays unset.
 const fetchSushiPool = async function ({
   pool,
   trackedAddresses,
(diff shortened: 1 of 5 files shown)


internal-dashboard/src/fetchers/fetchSushiPools.ts
```diff
@@ -105,15 +107,34 @@ const fetchSushiPool = async function ({
...
+      ...decimals,
+      price: Math.min(...ranges.map((range) => range.lowerPrice)),
+    }),
+    poolAddress: pool.address,
+    upperTick: priceToTick({
+      ...decimals,
+      price: Math.max(...ranges.map((range) => range.upperPrice)),
+    }),
+  });
+
```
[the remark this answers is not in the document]
gndelia: It's ok - this is just an internal dashboard

internal-dashboard/src/lib/v3PoolState.ts
```diff
@@ -0,0 +1,164 @@
...
+export const fetchV3PoolState = async function ({
+  lowerTick,
+  poolAddress,
+  upperTick,
+}: {
+  lowerTick: number;
+  poolAddress: Address;
+  upperTick: number;
+}) {
+  const [[sqrtPriceX96, currentTick], liquidity, spacing] = await multicall(
```
ArturDolzan, line 99: Might be worth a quick look at pinning these three reads to one `blockNumber` — they're sequential, and with `fallback()` over several public RPCs they could land on different heads, which the liquidity walk assumes they don't.


Artur's 🤖

internal-dashboard/src/lib/v3PositionMath.test.ts
```diff
@@ -81,7 +139,7 @@ describe("computeBandAmounts", function () {
 
   it("splits evenly for a price-symmetric band at $1 (equal decimals)", function () {
     const { amount0, amount1 } = computeBandAmounts({
-      decimals0: 18,
+      ...uniformPool,
```
ArturDolzan: Not sure, but maybe this one wants `currentTick: 0` alongside the other 18/18 overrides? It's inert while `uniformPool.initializedTicks` is empty, just looks like it could drift later.


Artur's 🤖

internal-dashboard/src/lib/v3PoolState.ts
```diff
@@ -0,0 +1,164 @@
...
+      args: [word] as const,
+      functionName: "tickBitmap" as const,
+    })),
+  });
+
+  const ticks = words.flatMap((word, index) =>
+    setBits(bitmaps[index]).map(
+      (bit) => (word * BITS_PER_WORD + bit) * spacing,
+    ),
+  );
```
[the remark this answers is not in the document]
gndelia: Fixed in [e4c41f15](https://github.com/vetro-protocol/vetro-monorepo/pull/706/commits/e4c41f15ec5c5bb9b40162ed6f110daba8241fa3).

Confirmed against the pool: 6 of the 15 ticks fetched for the configured band fell outside the requested window. The list is now filtered to `[min(lowerTick, currentTick), max(upperTick, currentTick)]` before the `ticks()` call, taking that batch from 15 to 9. `computeBandAmounts` only reads ticks inside that range, and a live read returns byte-identical amounts.

internal-dashboard/src/lib/v3PoolState.ts
gndelia: Fixed in [f6de1fc8](https://github.com/vetro-protocol/vetro-monorepo/pull/706/commits/f6de1fc89c9cdcf53c3d7455a375960c9855230a).

Good catch — the walk composes the current liquidity with the per-tick deltas, so those only add up at one head. The block number is now resolved once and all three multicalls are pinned to it. viem caches `getBlockNumber` for the polling interval, so the extra round trip is usually free.

One refinement on the mechanism: `fallback` defaults to `rank: false`, so the reads normally all hit transport 0 rather than spreading across providers. The likelier skew was plain block advancement across the ~600ms the three reads span. Pinning covers both. It also can't hard-fail against a lagging provider — "header not found" isn't in viem's `shouldThrow`, so `fallback` advances to the next transport.

internal-dashboard/src/lib/v3PositionMath.test.ts
gndelia: Fixed in [3995b1ed](https://github.com/vetro-protocol/vetro-monorepo/pull/706/commits/3995b1ede269fe82a001802866be0cbd417fa863).

Right on both counts — `currentTick` came from the 18/6 fixture while `sqrtPriceX96` was overridden to the 18/18 price, so they described different prices, and it was inert only because `initializedTicks` is empty there.

Rather than adding `currentTick: 0`, both fields now come from one price via the existing `atPrice` helper, so they can't drift apart again.

gabmontes (approved):
🤷🏼‍♂️

ArturDolzan (approved):

PR #711, opened 2026-08-11
Update target-yield-earn package

gndelia:
### Description

Follow-up of #708 - this PR gives `target-yield-earn` the same treatment as `earn`, `treasury`, `gateway` and `bridge`: it takes `isAddressValid` from the `core` package, switches relative imports to the `.ts` extension so consumers can import the package through its exports with no build step, and publishes through esbuild (which inlines the private `@vetro-protocol/core`).

### Screenshots

N/A

### Related issue(s)

Related to #659
Related to #695

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/target-yield-earn/package.json
+++ b/packages/target-yield-earn/package.json
@@ -17,11 +17,12 @@
   },
   "scripts": {
     "build": "tsc --noEmit",
-    "build:esm": "tsc -p tsconfig.build.json --outDir ./_esm --sourceMap",
+    "postbuild": "pnpm bundle && pnpm build:types",
     "build:types": "tsc -p tsconfig.build.json --declarationDir ./_types --emitDeclarationOnly --declaration --declarationMap",
+    "bundle": "node ../../scripts/bundle-package.js",
     "clean": "rm -rf ./_esm ./_types",
     "prepack": "cp ../../LICENSE LICENSE",
-    "prepublishOnly": "pnpm clean && pnpm build:esm && pnpm build:types",
+    "prepublishOnly": "pnpm clean && pnpm build",
     "test": "vitest run",
     "test:coverage": "vitest run --coverage"
   },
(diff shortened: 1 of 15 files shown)


packages/target-yield-earn/package.json
```diff
@@ -17,11 +17,12 @@
...
   "scripts": {
     "build": "tsc --noEmit",
-    "build:esm": "tsc -p tsconfig.build.json --outDir ./_esm --sourceMap",
+    "postbuild": "pnpm bundle && pnpm build:types",
     "build:types": "tsc -p tsconfig.build.json --declarationDir ./_types --emitDeclarationOnly --declaration --declarationMap",
+    "bundle": "node ../../scripts/bundle-package.js",
     "clean": "rm -rf ./_esm ./_types",
     "prepack": "cp ../../LICENSE LICENSE",
-    "prepublishOnly": "pnpm clean && pnpm build:esm && pnpm build:types",
+    "prepublishOnly": "pnpm clean && pnpm build",
```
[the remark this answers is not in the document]
gndelia: yes. This is temporary - it will eventually be published through that flow once contracts are defined. Because of that, I'm leaving the code ready for that

ArturDolzan (approved):

packages/target-yield-earn/package.json
```diff
@@ -17,11 +17,12 @@
   },
   "scripts": {
     "build": "tsc --noEmit",
```
gabmontes, line 19: This package uses the `postbuild` hook style (`prepublishOnly: clean && build`, with `postbuild` running `bundle && build:types`), matching `earn`. But `treasury`, `gateway` and `bridge` use the `prebundle` style (`prepublishOnly: clean && bundle && build:types`). Both produce the same output, but was following `earn` here intentional, or should we align all of them on one style to avoid drift?

— Auto Gabi

gndelia: This was fixed in parallel for those 3 packages in https://github.com/vetro-protocol/vetro-monorepo/pull/712

gabmontes (approved):

Issue #709, opened 2026-08-11
Add Variable yield pools details page


gndelia:
### Feature or change description

Just like VUSDx has the "Details page", the existing Earn pools will also have one.

PR #713, opened 2026-08-11
Extract InfoCard into a shared base component

gndelia:
### Description

The Borrow and Earn page had its own implementation of the same component, which is a visual card that shows some info + a label + optional subtitle. As the Target Yield page is coming and will use similar semantics to these pages (Plus Earn page will also gain their own details page), this PR unified that component into a shared one + stories.

Additionally, I compared it with Figma and fixed some paddings and spacings that were not consistent in both pages

#### Borrow

Mobile

image

Tablet

image

Desktop

image

#### Earn

Mobile

image

Tablet

image

Desktop

image

### Related issue(s)

Related to #709

### Checklist

- [x] Manual testing passed.


ArturDolzan (approved):

web/src/pages/earn/components/earnedAmountStat/index.tsx
```diff
@@ -1,20 +1,19 @@
...
-import { StatCard } from "../statCard";
 
 export function EarnedAmountStat() {
   const { t } = useTranslation();
-  const { data, isLoading } = useEarnedAmountUsd();
 
   return (
-    <StatCard
+    <InfoCard
+      {...useEarnedAmountUsd()}
```
gabmontes, line 13: Minor: prefer destructuring the two fields the card actually uses instead of spreading the whole `useQuery` result:

```tsx
const { data, isLoading } = useEarnedAmountUsd();
// ...
<InfoCard data={data} isLoading={isLoading} ... />
```

Spreading forwards ~15 unused query fields (`refetch`, `status`, `dataUpdatedAt`…) into a shared `base/` primitive. Harmless today, but if `InfoCard` ever gains a `...rest`→DOM passthrough these will spill onto the element. It also makes the data flow clearer. Same applies to `stakedBalanceStat`.

— Gabi 🤖

web/src/pages/earn/components/stakedBalanceStat/index.tsx
```diff
@@ -112,15 +112,14 @@ function FromPoolsBadge() {
 
 export function StakedBalanceStat() {
   const { t } = useTranslation();
-  const { data, isLoading } = useTotalStakedUsd();
 
   return (
-    <StatCard
-      badge={<FromPoolsBadge />}
+    <InfoCard
+      {...useTotalStakedUsd()}
```
gabmontes, line 118: Minor: same as `earnedAmountStat` — destructure `{ data, isLoading }` and pass them explicitly rather than spreading the full query result.

— Auto Gabi

web/src/components/borrow/marketInfoCards.tsx
```diff
@@ -62,88 +62,78 @@ const TrendIcon = () => (
...
           <InfoCard
-            badge={
-              <RenderCryptoValue
-                showSymbol
-                status={collateralStatus}
-                token={market.collateralToken}
-                value={collateralAssets}
-              />
-            }
+            {...collateral}
```
gabmontes, line 74: Minor: this mixes styles — `{...collateral}` is spread for `data`/`isLoading`, but the subtitle then reads `collateral.status`/`collateral.data` explicitly. Consider destructuring once and passing the fields explicitly for consistency.

— Gabi 🤖

web/src/pages/earn/components/earnedAmountStat/index.tsx
gndelia: Particularly with useQuery, it works a lil' better if we don't destructure - See [docs](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#tracked-properties)

Additionally, if those props end up in the DOM, warnings are logged in the terminal about unknown props, so it's detectable,.

web/src/pages/earn/components/stakedBalanceStat/index.tsx
gndelia: See https://github.com/vetro-protocol/vetro-monorepo/pull/713#discussion_r3766710805

web/src/components/borrow/marketInfoCards.tsx
gndelia: I can't forward directly with `{...collateral}` because the `RenderCryptoValue` expects `value` instead of `data`. Having said that, I don't think this is less readable at all, nor that destructuring is that essential. This is fine

gabmontes (approved):

Issue #694, opened 2026-08-05
Add missing E2E tests for the main flows


gndelia:
I think we should cover these scenarios:

## Swap flow

- [x] Cancelling a Redeem — #696

## Earn flow

- [x] Deleting an exit ticket (Cancel) — #707
- [x] Withdraw all #730

## Borrow flow

- [ ] Starting a position
- [ ] Borrowing more
- [ ] Supplying more
- [ ] Repay more
- [ ] Withdraw collateral

## Misc

- [x] 404 page #714

PR #714, opened 2026-08-12
Show the 404 page for unknown paths

gndelia:
### Description

Adds the E2E coverage for the 404 page (the last unchecked item under "Misc" in #694) — and fixes the routing bug that writing the test exposed.

An unsupported language prefix redirected to `/en` and threw away the rest of the path, so unknown paths never reached the 404 page at all. The blast radius was wider than just 404s:

| URL | Before | After |
| --- | --- | --- |
| `/borrwo` (typo) | `/en/swap` — Swap, no 404 | `/en/borrwo` — 404 |
| `/borrow` (valid page, no prefix) | `/en/swap` — **wrong page** | `/en/borrow` — Borrow |
| `/swap?amount=5` | `/en/swap` — query dropped | `/en/swap?amount=5` — kept |
| `/fr/swap` (unsupported language) | `/en` — Swap | `/en/fr/swap` — 404 |
| `/`, `/en`, `/es/earn`, `/en/borrwo` | unchanged | unchanged |

Any unprefixed valid path (`/borrow`, `/earn`, `/analytics`) silently landed on Swap; only `/swap` appeared to work, coincidentally, because Swap is the index redirect target. Query strings were dropped on those paths too, which matters since the Swap form keeps state in the URL via nuqs.

The fix prefixes the fallback language onto the current path instead of replacing it, preserving search and hash. The redirect had to move out of `I18nInitializer`'s layout effect and into `LanguageRoutes`' render: the language routes include an index route that navigates to Swap, and it wins the race if both redirects run from effects (the first attempt produced `/en/this-page-does-not-exist/swap`). That constraint is the one comment left on the new hook.

Two deliberate behavior changes worth calling out for review: an unsupported language now 404s rather than falling back to English Swap, and that includes region-tagged prefixes like `/en-US/swap`.

### Screenshots

N/A — no new UI. The `NotFound` page already existed; this only changes which URLs route to it.

### Related issue(s)

Related to #694

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/e2e/notFound.spec.ts
+++ b/web/e2e/notFound.spec.ts
@@ -1,0 +1,53 @@
+import { expect, test } from "@playwright/test";
+
+const notFoundTitle = "Page not found";
+const notFoundDescription = "The page you're trying to access does not exist.";
+
+test("renders the 404 page on an unknown path", async function ({ page }) {
+  await page.goto("/en/this-page-does-not-exist");
+
+  await expect(
+    page.getByRole("heading", { name: notFoundTitle }),
+  ).toBeVisible();
+  await expect(page.getByText(notFoundDescription)).toBeVisible();
+  await expect(page).toHaveURL(/\/en\/this-page-does-not-exist$/);
+});
+
+test("renders the 404 page on an unknown path with no language prefix", async function ({
+  page,
+}) {
+  await page.goto("/this-page-does-not-exist");
+
+  await expect(
+    page.getByRole("heading", { name: notFoundTitle }),
+  ).toBeVisible();
+  await expect(page).toHaveURL(/\/en\/this-page-does-not-exist$/);
+});
+
+test("renders the 404 page on an unsupported language", async function ({
+  page,
+}) {
+  await page.goto("/fr/swap");
+
+  await expect(
+    page.getByRole("heading", { name: notFoundTitle }),
+  ).toBeVisible();
+  await expect(page).toHaveURL(/\/en\/fr\/swap$/);
+});
+
+test("the root path redirects to swap", async function ({ page }) {
+  await page.goto("/");
+
+  await expect(page).toHaveURL(/\/en\/swap$/);
+  await expect(
+    page.getByRole("heading", { name: "Swap your tokens with Vetro assets" }),
+  ).toBeVisible();
+});
+
+test("the 404 page links back to swap", async function ({ page }) {
+  await page.goto("/en/this-page-does-not-exist");
+
+  await page.getByRole("link", { name: "Take me back to swap" }).click();
+
+  await expect(page).toHaveURL(/\/en\/swap$/);
+});
--- a/web/e2e/notFound.spec.ts
+++ b/web/e2e/notFound.spec.ts
@@ -26,6 +26,30 @@ test("renders the 404 page on an unknown path with no language prefix", async fu
   );
 });
 
+test("renders the requested page when the path has no language prefix", async function ({
+  page,
+}) {
+  await page.goto("/contact");
+
+  await expect(page).toHaveURL(/\/en\/contact$/);
+  await expect(
+    page.getByRole("heading", {
+      name: "Tell us what happened so we can help",
+    }),
+  ).toBeVisible();
+});
+
+test("preserves a deep link when the path has no language prefix", async function ({
+  page,
+}) {
+  const marketId =
+    "0x7d1306d23f9f1e419697b8275001db9ea74b3c75190a7db8f5d81fed2fb94561";
+
+  await page.goto(`/borrow/${marketId}`);
+
+  await expect(page).toHaveURL(new RegExp(`/en/borrow/${marketId}$`));
+});
+
 test("renders the 404 page on an unsupported language", async function ({
   page,
 }) {
--- a/web/e2e/notFound.spec.ts
+++ b/web/e2e/notFound.spec.ts
@@ -16,12 +16,14 @@ test("renders the 404 page on an unknown path", async function ({ page }) {
 test("renders the 404 page on an unknown path with no language prefix", async function ({
   page,
 }) {
-  await page.goto("/this-page-does-not-exist");
+  await page.goto("/this-page-does-not-exist?foo=bar#section");
 
   await expect(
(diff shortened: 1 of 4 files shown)


web/e2e/notFound.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+test("renders the 404 page on an unknown path with no language prefix", async function ({
+  page,
+}) {
+  await page.goto("/this-page-does-not-exist");
+
+  await expect(
+    page.getByRole("heading", { name: notFoundTitle }),
+  ).toBeVisible();
+  await expect(page).toHaveURL(/\/en\/this-page-does-not-exist$/);
```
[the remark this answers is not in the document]
gndelia: Fixed in [600a6c16](https://github.com/vetro-protocol/vetro-monorepo/pull/714/commits/600a6c1662af683d7d22563e35be55176f0cad28).

Good catch — none of the five tests used a `?` or `#`, so dropping `${search}${hash}` from the redirect kept the whole suite green. The no-prefix test now navigates to `/this-page-does-not-exist?foo=bar#section` and asserts both survive. I confirmed the assertion fails when the hook drops them, so it isn't a test that only passes.

web/e2e/notFound.spec.ts
gabmontes, line 1: This suite covers the 404 and root-redirect cases, but not the headline routing fix: an unprefixed *valid* page now resolving correctly (e.g. `/borrow` → `/en/borrow` rendering Borrow, deep links like `/borrow/123` too). A test asserting that would lock in the actual regression this PR repairs.

Related: the fix makes this comment in `web/e2e/earn.spec.ts` (~L74) stale — it now describes the old buggy behavior:
`// a bare page.goto("/earn") is parsed as lang="earn" and falls back to the default page`
After this change `/earn` redirects to `/en/earn` and renders Earn, so that navigation could be simplified to `page.goto("/en/earn")` and the comment updated. Worth fixing here to avoid leaving a misleading note behind.

— Gabi 🤖

web/src/i18n/config.tsx
```diff
@@ -17,22 +17,31 @@ export const initializeI18n = () =>
...
+ * route that navigates to Swap, and it wins if both redirects run from effects.
+ */
+export const useUnsupportedLanguageRedirect = function () {
+  const { lang } = useParams<{ lang: string }>();
+  const { hash, pathname, search } = useLocation();
+
+  if (!lang || supportedLngs.includes(lang)) {
+    return undefined;
+  }
+  return `/${fallbackLng}${pathname}${search}${hash}`;
```
gabmontes, line 33: Confirming intent: any unsupported prefix now hard-404s, including region-tagged variants of supported languages — `/en-US/swap` redirects to `/en/en-US/swap` and lands on NotFound rather than English Swap. That's a reasonable, documented choice given there's no `Accept-Language` detection. Only flagging in case a graceful fallback to the base language (`en-US` → `en`, `es-MX` → `es`) is preferable to a 404 for those region variants.

— Auto Gabi

gndelia:
There's an Earn test failing with a reverting contract, unrelated to this PR 🤷🏽

web/src/i18n/config.tsx
gndelia: yes, it's deliberate - mentioned in the PR description: 

> Two deliberate behavior changes worth calling out for review: an unsupported language now 404s rather than falling back to English Swap, and that includes region-tagged prefixes like /en-US/swap.

web/src/i18n/config.tsx
```diff
@@ -17,22 +17,31 @@ export const initializeI18n = () =>
     supportedLngs,
   });
 
+/**
+ * Callers must redirect during render — the language routes include an index
+ * route that navigates to Swap, and it wins if both redirects run from effects.
+ */
```
ArturDolzan, line 25: Perhaps it's worth rewording this note slightly — `<Navigate>` also fires from an effect in react-router 8, so what actually saves us is returning early so the index route never mounts. Might read clearer for whoever touches this next.

web/e2e/notFound.spec.ts
gndelia: Fixed in [b8a4f30b](https://github.com/vetro-protocol/vetro-monorepo/pull/714/commits/b8a4f30b63d2e75f5dff927b6cd5fdeb8441668b) and [b0635c84](https://github.com/vetro-protocol/vetro-monorepo/pull/714/commits/b0635c84e05f75fd7b84601bd08269c9f6cd0eb9).

(Apologies for the edit — the first version of this reply went out with a placeholder body by mistake.)

Both points were right. On coverage: `swap.spec.ts` does `page.goto("/swap")`, but that passed before the fix too (`/swap` → `/en` → index `<Navigate to="swap">` → `/en/swap`), so nothing in the suite actually locked in the routing repair. Added two tests: `/contact` → `/en/contact` rendering the Contact page, and `/borrow/0x7d13…` → `/en/borrow/0x7d13…` for the multi-segment deep link. I used Contact rather than Borrow for the render assertion because its heading is a plain `<h1>`, whereas the Borrow title waits on `usePeggedToken` — same routing assertion, no chain dependency. The deep-link test asserts the URL only, for the same reason. I confirmed both fail when the redirect drops the path, so neither is a test that only passes.

On the stale comment: correct, `/earn` now redirects to `/en/earn` and renders Earn. I dropped the outdated explanation and kept the nav click, since going through the navbar is still what a user does.

web/src/i18n/config.tsx
gndelia: Fixed in [13859352](https://github.com/vetro-protocol/vetro-monorepo/pull/714/commits/1385935293f812d988350ed5b9a8340775ee802c).

You are right — I checked `react-router@8.3.0` (`dist/development/lib/components.js:380-402`) and `Navigate` does call `navigate()` inside a `useEffect`, so "redirect during render" was describing something that does not happen. The note now says what actually holds: the caller returns early, so the index route never mounts and never schedules its redirect to Swap.

gndelia:
> There's an Earn test failing with a reverting contract, unrelated to this PR 🤷🏽

no longer failing 😅

ArturDolzan (approved):
LGTM!

gabmontes (approved):

PR #718, opened 2026-08-13
Catch package changes that break dependents

gndelia:
### Description

Follow-up to #717, which broke CI in a way the local checks couldn't catch: a change in `@vetro-protocol/core` passed the package's own build but failed treasury's. Packages are consumed as source, so each dependent recompiles them under its own tsconfig, where a narrower `lib`/`types` set rejects code the package's own build accepts — in that case an ambient `URL` global.

- The `run-checks` skill's scoped recipe now builds the changed package **and its dependents**, which is the only step that compiles a package under a consumer's tsconfig. It also covers the apps, whose Vite bundle catches broken `exports` maps and unresolvable imports that `tsc` can't see. Three stale "(= `tsc --noEmit`)" glosses are gone too, since a package `build` now bundles and emits declarations.
- `packages/core` had no `build:types`, so it never typechecked its own `src` the way its consumers do — its own build sees `@types/node` (pulled in via `vitest.config.ts`) while consumers' builds don't. It now uses the same `tsconfig.build.json` and bundle pipeline as the other packages.

Verified by dropping a `URL.canParse` line into `packages/core/src` and confirming the new command reproduces the same `TS2552` that CI reported.

### Screenshots

N/A — tooling only.

### Related issue(s)

No issue — follow-up to #717.

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/.claude/skills/run-checks/SKILL.md
+++ b/.claude/skills/run-checks/SKILL.md
@@ -20,13 +20,20 @@ a filter (`...<pkg>`) adds dependents, so a change in a shared package still
 re-tests `web`/`api`. (A _trailing_ `...` would instead pull in the package's
 dependencies — the wrong direction here.)
 
-- Changed a **package** (`packages/*`) — its typecheck _is_ its `build`
-  (`tsc --noEmit`, no bundle):
+- Changed a **package** (`packages/*`) — its typecheck _is_ its `build`:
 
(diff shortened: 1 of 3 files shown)


ArturDolzan (approved):
uACK

gabmontes (approved):

PR #719, opened 2026-08-13
List target fixed yield pools on Earn

gndelia:
### Description

Adds the pools list under "Target fixed yield pools" on Earn, behind the fixed-term-yield flag: Token, Pool contract, Total deposits against the epoch cap, Target fixed APY, Term state and Your staked amount.

The contracts don't exist yet, so every value is mocked at the deepest layer — the fetchers under `web/src/fetchers/earn/targetYieldPool/` return hardcoded data keyed by epoch id, and `targetYieldVaults.ts` points the on-chain reads at the VUSD staking vault while the row's own address stays `zeroAddress`. Components and hooks are written as if the data were real, so each swap is a local change to one fetcher, and every TODO names the contract call that replaces it. `VUSDx` initializes its ERC-4626 asset with VUSD and mints ERC-20 shares, so `asset()` and the share balance already behave the same — the address is the only thing that changes for those two.

`getTermState` is deliberately left with a TODO: on the vault the entry window spans the whole epoch while the exit window is only its last 1–7 days, so the two overlap and the real mapping still needs confirming.

`PoolInfoItem` is reworked so both pool rows share it — it now takes a query's `data`, `isPending` and `isError` plus an optional `render`, instead of a pre-formatted `value` and `children`. `isPending` rather than `isLoading` because a query disabled while waiting on the epoch reports `isLoading` false, which rendered "-" before the data arrived.

**Commits:**

a2d71da2 Drive PoolInfoItem from query state
920d894b List target fixed yield pools on Earn

### Screenshots

image

### Related issue(s)

Related to #646

### Checklist

- [x] Manual testing passed.


web/src/pages/earn/hooks/targetYieldPool/useEpochId.ts
```diff
@@ -0,0 +1,18 @@
+import { queryOptions, useQuery } from "@tanstack/react-query";
+import { fetchEpochId } from "fetchers/earn/targetYieldPool/fetchEpochId";
+
+import { targetYieldVaultReadAddress } from "../../targetYieldVaults";
+
+const epochIdOptions = () =>
+  queryOptions({
+    queryFn: fetchEpochId,
+    queryKey: ["target-yield-epoch-id", targetYieldVaultReadAddress],
```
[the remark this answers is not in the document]
gndelia: Right - so polling will be added through `refetchInterval` once we can query epochId and access `end` - this way we know the epoch has ended and we may poll for a new one

web/src/pages/earn/components/fixedTermEarn/poolInfoBar.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <div className="flex flex-col gap-6 border-b border-gray-200 bg-white p-4 sm:gap-4 md:flex-row md:items-center md:px-16 md:py-6">
+      <div className="grid min-w-0 grid-cols-2 gap-4 sm:flex sm:items-center sm:justify-center-safe sm:gap-6 sm:overflow-x-auto md:justify-start md:gap-8">
+        <PoolToken peggedToken={peggedToken} />
+        <PoolContract address={vaultAddress} />
+        <PoolDeposits />
+        <PoolTargetApy />
+        <PoolTermState />
+        <PoolStakedAmount />
```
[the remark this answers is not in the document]
gndelia: Fixed in [9fc531a5](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/9fc531a534117957f8ab31e069cd20c6dbb48c2f). The row address is now a parameter to every data hook, and `targetYieldVaultAddresses` points at the VUSD staking vault instead of the zero address so those reads resolve until the real vaults are deployed.

web/src/pages/earn/hooks/usePoolStakedAmount.ts
```diff
@@ -0,0 +1,59 @@
...
+  account: Address | undefined;
+  chainId: number;
+  client: Client | undefined;
+  stakingVaultAddress: Address;
+}) =>
+  queryOptions({
+    enabled: !!client && !!account,
+    async queryFn({ client: queryClient }) {
+      const [peggedToken, stakedBalance] = await Promise.all([
+        queryClient.ensureQueryData(
```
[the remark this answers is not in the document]
gndelia: Fixed in [94917095](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/94917095470e1f8ba1bb60ae35063dafaab9c0aa). Confirmed the staleness first — six mutations (`useStakeDeposit`, `useStakeWithdraw`, `useInstantWithdraw`, `useClaimWithdraw`, `useClaimWithdrawBatch`, `useCancelWithdraw`) patch `stakedBalanceQueryKey` with `setQueryData`, which the derived copy never saw. The hook now observes both queries with `useQueries` instead. Verified against a mainnet fork: depositing 10 VUSD moves the row from $0.00 to $9.99 with no reload.

web/src/pages/earn/hooks/targetYieldPool/useDeposits.ts
```diff
@@ -0,0 +1,39 @@
...
+    queryFn: ({ client: queryClient }) =>
+      fetchDeposits({
+        client: client!,
+        epochId: epochId!,
+        queryClient,
+        stakingVaultAddress,
+      }),
+    queryKey: [
+      "target-yield-pool-deposits",
+      client?.chain?.id,
```
gabmontes: `useDeposits` keys its query on `client?.chain?.id`, but the sibling epoch-scoped reads (`useEpochId`, `useTargetApy`, `useTerm`) don't — even though all four become on-chain reads and the rest of the earn hooks (`poolDepositsQueryKey`, `vaultPeggedTokenQueryKey`, `useCooldownDuration`) consistently key on chain id. Is the chain id here intentional? Since the app is currently pinned to Ethereum mainnet (`useMainnet` / `useEthereumClient`), it makes no functional difference today, so either it's redundant and can be dropped from `useDeposits`, or it's the intended convention and the three siblings should match it for when the network selector lands. Which direction do you prefer?

— Gabi 🤖

web/src/pages/earn/components/poolInfoBar/poolInfoItem.tsx
```diff
@@ -1,32 +1,39 @@
...
-type Props = {
-  children?: ReactNode;
-  isLoading?: boolean;
+type Props<T> = {
+  data: T | undefined;
+  isError?: boolean;
+  isPending?: boolean;
   label: string;
-  value?: string;
+  render?: (data: T) => ReactNode;
```
ArturDolzan, line 9: Perhaps it's worth a quick look at `web/stories/poolInfoItem.stories.tsx`? It still passes `value` / `isLoading` / `children`, and since `stories` isn't in the tsconfig `include` and Storybook builds transpile-only, I think all four stories would just render `-` now.

web/src/pages/earn/components/fixedTermEarn/poolStakedAmount.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useAccount } from "wagmi";
+
+import { usePoolStakedAmount } from "../../hooks/usePoolStakedAmount";
+import { PoolInfoItem } from "../poolInfoBar/poolInfoItem";
+
+type Props = {
+  stakingVaultAddress: Address;
+};
+
+export function PoolStakedAmount({ stakingVaultAddress }: Props) {
```
ArturDolzan: Might be worth confirming whether `fixedTermEarn/poolStakedAmount.tsx` needs to exist. It looks identical to `poolInfoBar/poolInfoStakedAmount.tsx` apart from the name, and the row already pulls `PoolContract` / `PoolToken` from that folder.

web/src/pages/earn/components/poolInfoBar/index.tsx
```diff
@@ -49,20 +49,22 @@ export function PoolInfoBar({ stakingVaultAddress }: Props) {
...
-            </span>
-          </div>
-        </PoolInfoItem>
+          render={(formattedApy) => (
+            <div className="flex items-center gap-1">
+              <span className="text-xsm font-semibold text-gray-900">
+                {formattedApy}
+              </span>
+            </div>
+          )}
```
ArturDolzan: Not sure, but maybe the APY `render` can go now? It looks like it produces the same span as the default, and the `flex items-center gap-1` wrapper has been holding a single child since the icon was dropped.

web/src/pages/earn/components/fixedTermEarn/poolStakedAmount.tsx
gndelia: Yikes - I don't know how this isn't caught by CI. I'll take a look as well

gndelia: I will fix this particular case, but I created an issue https://github.com/vetro-protocol/vetro-monorepo/issues/725 to ensure CI picks this problem when it happens again

gndelia: ok so the previous 2 comments were meant to go here https://github.com/vetro-protocol/vetro-monorepo/pull/719#discussion_r3784137044

I guess IA doesn't mess this up 😓

web/src/pages/earn/components/poolInfoBar/poolInfoItem.tsx
gndelia: See https://github.com/vetro-protocol/vetro-monorepo/pull/719#discussion_r3784218663 - I commented in the wrong discussion lol

web/src/pages/earn/hooks/targetYieldPool/useDeposits.ts
gndelia: Dropped it in [aa423c24](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/aa423c24ba91a95b05cae7f4ad633e1c6b3d4582), so the four target yield pool keys now match. The three sibling hooks take no client today — their fetchers are hardcoded mocks with a TODO to read from `@vetro-protocol/target-yield-earn` — so a chain id there would name an input the query does not read. When the real reads land, each fetcher gets a client and the chain id enters every key at that point.

web/src/pages/earn/components/poolInfoBar/poolInfoItem.tsx
gndelia: Good catch. Fixed in [d7c7568d](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/d7c7568de23e29c59cc370244530abb8dad5dc04) — the four stories now use `data` / `isPending` / `render`. You were right that `tsc` cannot see them: `stories` is not in the `web/tsconfig.json` include. Adding it exposes 6 other story files with the same drift, so that cleanup goes to its own issue.

web/src/pages/earn/components/fixedTermEarn/poolStakedAmount.tsx
gndelia: Fixed in [5e2008d3](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/5e2008d37c147878b03c7e0f8f0e879e7ee8b6db). The duplicate is deleted and the fixed term bar now uses `PoolInfoStakedAmount`, like it already does for `PoolContract` and `PoolToken`.

web/src/pages/earn/components/poolInfoBar/index.tsx
gndelia: Right, it can. Removed in [1e5e2b3b](https://github.com/vetro-protocol/vetro-monorepo/pull/719/commits/1e5e2b3b4de3d7f7cb4e66408ac23df6d165e045). The default render makes the same span, and this matches `PoolTargetApy` on the fixed term bar.

ArturDolzan (approved):

gabmontes (approved):

PR #729, opened 2026-08-14
Include Storybook in the web typecheck

### LINKED ISSUE #725: Include Storybook in web builds

We were not including storybook in the compliation process. 

This may cause that some stories stop working if the component's prop changes. The build process would detect this so we can fix them.

### Proposed implementation

Include it in the compilation of `web` workspace, and fix the stories that are broken

gndelia:
### Description

The `web` workspace did not compile its Storybook stories. `web/tsconfig.json` left `stories` and `.storybook` out of `include`, so `tsc` never saw them, and a component prop change could break a story with nothing to report it.

This adds both to the include list. `pnpm --filter web build` runs `tsc && vite build`, so the stories are now covered by the build and by the `JS Checks` CI job.

Six stories had already drifted from their components and are fixed here:

- `activityList` — pass the required `hasTransactions`
- `allocationLegend` — set `formatAmount` in the meta args, using the same `formatUsd` default that `AllocationCard` applies
- `modal` — `children` is now a `({ close }) => ReactNode` render prop
- `searchInput` — move the required `value` and `onChange` into the meta args
- `segmentedControl` — seed the state from `tokenSymbols[0]`, which is typed as `TokenSymbol`
- `shadow` — use a type-only import for `ComponentProps`

A seventh, `fiatValue`, was broken at runtime in a way the typecheck does not catch: all three of its stories rendered `-`. It had no `WagmiProvider`, so `usePublicClient` threw, and the component's own `<ErrorBoundary fallback="-">` turned that crash into the fallback. It also seeded the query key `["token-price"]`, while `usePrices` reads `["prices", chainId]`. Both are fixed, and the three stories now render `125.50`, a skeleton, and `-` respectively.

### Screenshots

No application UI changes — the only visual changes are in Storybook. Every changed story was checked in a real browser: `tsc`, `eslint`, `prettier --check` pass, `storybook:build` completes, and no story logs a console error.

### Related issue(s)

Closes #725

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/stories/fiatValue.stories.tsx
+++ b/web/stories/fiatValue.stories.tsx
@@ -2,8 +2,10 @@ import type { Meta, StoryObj } from "@storybook/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import type { Token } from "@vetro-protocol/core";
 import { parseUnits } from "viem";
+import { createConfig, http, WagmiProvider } from "wagmi";
 
 import { RenderFiatValue } from "../src/components/base/fiatValue";
+import { mainnet } from "../src/networks/mainnet";
 
 const usdc: Token = {
   address: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
@@ -14,22 +16,36 @@ const usdc: Token = {
   symbol: "USDC",
 };
 
-const createQueryClient = function (prices?: Record<string, string>) {
-  const client = new QueryClient({
+const wagmiConfig = createConfig({
+  chains: [mainnet],
+  transports: { [mainnet.id]: http() },
+});
+
+const createQueryClient = function () {
+  const queryClient = new QueryClient({
     defaultOptions: {
       queries: {
         retry: false,
       },
     },
   });
-  if (prices) {
-    client.setQueryData(["token-price"], prices);
-  }
-  return client;
+  queryClient.setQueryData(["prices", mainnet.id], { USDC: "1.00" });
+  return queryClient;
 };
+
+const queryClient = createQueryClient();
 
 const meta = {
   component: RenderFiatValue,
+  decorators: [
+    (Story) => (
+      <WagmiProvider config={wagmiConfig}>
+        <QueryClientProvider client={queryClient}>
+          <Story />
+        </QueryClientProvider>
+      </WagmiProvider>
+    ),
+  ],
   title: "Components/FiatValue",
 } satisfies Meta<typeof RenderFiatValue>;
 
@@ -43,13 +59,6 @@ export const WithValue: StoryType = {
     token: usdc,
     value: parseUnits("125.50", usdc.decimals),
   },
-  decorators: [
-    (Story) => (
-      <QueryClientProvider client={createQueryClient({ USDC: "1.00" })}>
-        <Story />
-      </QueryClientProvider>
-    ),
-  ],
 };
 
 export const Loading: StoryType = {
@@ -58,13 +67,6 @@ export const Loading: StoryType = {
     token: usdc,
     value: undefined,
   },
-  decorators: [
-    (Story) => (
-      <QueryClientProvider client={createQueryClient()}>
-        <Story />
-      </QueryClientProvider>
-    ),
-  ],
 };
 
 export const Error: StoryType = {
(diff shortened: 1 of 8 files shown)


gabmontes (dismissed):

web/stories/fiatValue.stories.tsx
```diff
@@ -14,22 +16,36 @@ const usdc: Token = {
...
       queries: {
         retry: false,
       },
     },
   });
-  if (prices) {
-    client.setQueryData(["token-price"], prices);
-  }
-  return client;
+  queryClient.setQueryData(["prices", mainnet.id], { USDC: "1.00" });
```
ArturDolzan: Perhaps it's worth checking that `pricesOptions` hardcodes `refetchInterval: 60_000`, so after a minute these stories fire a real portal/RPC fetch. If it fails, `Loading` would flip to `-` instead of the skeleton.

web/stories/fiatValue.stories.tsx
```diff
@@ -14,22 +16,36 @@ const usdc: Token = {
...
 };
 
-const createQueryClient = function (prices?: Record<string, string>) {
-  const client = new QueryClient({
+const wagmiConfig = createConfig({
+  chains: [mainnet],
+  transports: { [mainnet.id]: http() },
+});
+
+const createQueryClient = function () {
```
ArturDolzan: Not sure, but maybe `createQueryClient` could just be inlined now that it takes no args and is called once right below? Would save a level of indirection.

web/stories/fiatValue.stories.tsx
gndelia: This is true. I don't like any of the alternatives, and this is largely pre-existing I think, so I'll create an issue to solve it. In the meantime, the story can be used at least for a minute 😆

gndelia: https://github.com/vetro-protocol/vetro-monorepo/issues/731 created

web/stories/fiatValue.stories.tsx
gndelia: Fixed in [e54bab7a](https://github.com/vetro-protocol/vetro-monorepo/pull/729/commits/e54bab7a6bac4571f4f3a80d889c8897e199f7cb).

ArturDolzan (approved):

web/stories/fiatValue.stories.tsx
```diff
@@ -58,13 +62,6 @@ export const Loading: StoryType = {
     token: usdc,
     value: undefined,
```
gabmontes, line 63: Nit: the Loading story's Skeleton renders because `value` is `undefined` (branch 3, the catch-all), not because of `queryStatus: "pending"` — the component only checks `queryStatus` for the `"error"` case. So the `"pending"` here is cosmetic. Fine to keep for intent, just flagging it doesn't actually drive the loading state. — Gabi 🤖

web/stories/fiatValue.stories.tsx
```diff
@@ -14,22 +16,31 @@ const usdc: Token = {
...
-  }
-  return client;
-};
+  },
+});
+queryClient.setQueryData(["prices", mainnet.id], { USDC: "1.00" });
 
 const meta = {
   component: RenderFiatValue,
+  decorators: [
```
gabmontes, line 35: Note: only `WithValue` actually reads the prices query — `Loading` and `Error` short-circuit on `value === undefined` before `pricesData` is used, so the `WagmiProvider` + prefilled client aren't exercised by them. Meta-level placement is still the right call (cleaner than per-story), just noting these two stories don't validate the fiat-price path. — Auto Gabi

web/stories/fiatValue.stories.tsx
gndelia: True, but that's how `react-query` derives status for "Loading": If the value's not defined and there's no error,  it must be in loading state. While we could remove `queryStatus`, let's leave it for intent

web/stories/fiatValue.stories.tsx
gndelia: I mean, the component internally uses `usePrices` which can't be called conditionally (it's a hook), so it's needed in the setup. Yeah, maybe some of these stories don't use the output of that hook, but I think they are needed for the setup because of the unconditionality of that call.

gabmontes (approved):

PR #730, opened 2026-08-14
Add an e2e test for Earn's Withdraw all

gndelia:
### Description

Adds the missing e2e coverage for "Withdraw all" on Earn — the last open Earn item in #694.

The test seeds four exit tickets (two on the VUSD pool, one on vetBTC, one left in cooldown), clicks "Withdraw all", and asserts the batching the feature exists for: one claim transaction per pool rather than one per ticket, the exact payout per pool, the resulting on-chain balances, and the cooldown ticket left untouched.

Writing it surfaced a real bug, fixed in the first commit. Each staking vault numbers its withdraw requests from its own counter, so a request id only identifies a ticket together with its vault. sVUSD is at request 20 and sVetBTC at 7 today, so the ids do not overlap — but nothing keeps them apart. Once they do, the exit tickets table gives distinct rows the same React key, and a claim on one vault marks an unclaimed ticket on the other as withdrawn. The table row id and the optimistic cache updates for claim, batch claim and cancel now match on the vault address as well.

To reach two pools, the seeding helpers take the token and the vault instead of hardcoding VUSD. `requestWithdraw` also steps the fork clock itself: two writes on the same vault one block apart get their gas estimated before the yield drip turns non-zero, and then run out of gas.

**Commits:**

1ff92914 Key exit tickets by vault and request id
56a1ec5a Add an e2e test for Earn's Withdraw all

### Screenshots

https://github.com/user-attachments/assets/78af3f2b-33cb-47d9-8178-fc1aaef7b19a

### Related issue(s)

Related to #694

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/e2e/earn.spec.ts
+++ b/web/e2e/earn.spec.ts
@@ -769,11 +769,11 @@ test("withdraw all the ready exit tickets", async function ({
 }) {
   const publicClient = createEthereumClient();
 
-  const VETBTC_DEPOSIT_AMOUNT = parseUnits("2", vetBtc.decimals);
-  const VETBTC_TICKET_AMOUNT = parseUnits("1", vetBtc.decimals);
-  const FIRST_VUSD_TICKET_AMOUNT = parseUnits("2", vusd.decimals);
-  const SECOND_VUSD_TICKET_AMOUNT = parseUnits("1", vusd.decimals);
-  const COOLDOWN_VUSD_TICKET_AMOUNT = parseUnits("1", vusd.decimals);
+  const vetBtcDepositAmount = parseUnits("2", vetBtc.decimals);
+  const vetBtcTicketAmount = parseUnits("1", vetBtc.decimals);
+  const firstVusdTicketAmount = parseUnits("2", vusd.decimals);
+  const secondVusdTicketAmount = parseUnits("1", vusd.decimals);
+  const cooldownVusdTicketAmount = parseUnits("1", vusd.decimals);
 
   // Seed a staked position in both pools, with cooldown on and no
   // instant-withdraw whitelist so every request lands in the queue as a ticket.
@@ -783,7 +783,7 @@ test("withdraw all the ready exit tickets", async function ({
     vaultAddress: sVusdAddress,
   });
   await stake({
-    assets: VETBTC_DEPOSIT_AMOUNT,
+    assets: vetBtcDepositAmount,
     token: vetBtc.address,
     vaultAddress: sVetBtcAddress,
   });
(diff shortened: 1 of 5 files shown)


web/e2e/earn.spec.ts
```diff
@@ -722,3 +762,235 @@ test("delete an exit ticket while it is in cooldown", async function ({
...
+  page,
+  walletTxHashes,
+}) {
+  const publicClient = createEthereumClient();
+
+  const VETBTC_DEPOSIT_AMOUNT = parseUnits("2", vetBtc.decimals);
+  const VETBTC_TICKET_AMOUNT = parseUnits("1", vetBtc.decimals);
+  const FIRST_VUSD_TICKET_AMOUNT = parseUnits("2", vusd.decimals);
+  const SECOND_VUSD_TICKET_AMOUNT = parseUnits("1", vusd.decimals);
+  const COOLDOWN_VUSD_TICKET_AMOUNT = parseUnits("1", vusd.decimals);
```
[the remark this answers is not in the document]
gndelia: Fixed in [5b88ddec](https://github.com/vetro-protocol/vetro-monorepo/pull/730/commits/5b88ddeca597c1613530761e3be34654f7cb7050). Renamed all five to camelCase. UPPER_SNAKE is the convention for the module-level spec constants, but these sit inside the test body, and nothing else in `web/` names a function-scoped constant that way.

ArturDolzan (approved):

web/e2e/earn.spec.ts
```diff
@@ -722,3 +762,233 @@ test("delete an exit ticket while it is in cooldown", async function ({
...
+    requestTxHash: request.requestTxHash,
+    shares: request.shares.toString(),
+    stakingVaultAddress,
+  });
+
+  let exitTicketsBody: ExitTicket[] = [
+    toTicket({
+      claimableAt: "1",
+      request: firstVusdRequest,
+      stakingVaultAddress: sVusdAddress,
```
gabmontes, line 875: The mocked tickets use the checksummed vault constants, but the real API returns **lowercase** addresses: `getUserExitTickets` returns the subgraph response verbatim (no `checksumAddress`, unlike the position handler in the same file) and the route does `c.json(data)` with no normalization.

This test guards vault-scoped ticket identity — the exact thing this PR fixes — but it exercises a casing the API never emits. It passes today only because the hooks use `isAddressEqual`; if a raw `===`/string comparison against a vault constant were reintroduced (the bug class this PR addresses), the checksummed mock would still pass and hide it.

Suggest lowercasing the mock `stakingVaultAddress` values (e.g. `sVusdAddress.toLowerCase()`) so the fixture mirrors the real API shape and the test actually covers the casing path.

— Auto Gabi

gndelia: I actually think what we should fix is `getUserExitTickets` because the IA instructions are configured to use `isAddressEqual` - so I think it's unlikely that we use `===` to compare addresses. Still, I'll create an issue to make that update in the API

gndelia: https://github.com/vetro-protocol/vetro-monorepo/issues/736

gabmontes (approved):

PR #733, opened 2026-08-17
Document internal dashboard env vars

### LINKED ISSUE #732: Allow Internal dashboard to configure RPC url with env variable

We want the internal dashboard to be able to configure the RPC url through Wrangler + env variable. This is because as this site is private (Behind CF), we can use a paid RPC for better performance. Locally, we can continue to use the public RPCs intended for development

gndelia:
### Description

The dashboard already reads `VITE_RPC_URL_MAINNET` at build time, for the viem client and for the Worker `connect-src` alike, so a paid RPC needs no code change: set the variable in the Cloudflare Workers Builds variables and the build picks it up over the committed `.env`. This PR documents that, marks the variable as required (unset, viem's default endpoint is blocked by the CSP) and corrects the stale tab list.

### Screenshots

N/A — no UI changes.

### Related issue(s)

Closes #732

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/internal-dashboard/README.md
+++ b/internal-dashboard/README.md
@@ -15,7 +15,7 @@ that serves the static assets and injects security headers (see `src/index.ts`).
 
 ## Environment Variables
 
-Set these in `internal-dashboard/.env` (or a `.env.local` override) before you run `dev` or `build`.
+Set these in `internal-dashboard/.env` (or a `.env.local` override) before you run `dev` or `build`. The committed `.env` holds the public development values only.
 
 | Variable               | Required | Description                                                                                                                                                          |
 | ---------------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
@@ -23,6 +23,8 @@ Set these in `internal-dashboard/.env` (or a `.env.local` override) before you r
 | `VITE_RPC_URL_MAINNET` | Yes      | RPC URL(s) for Ethereum mainnet. Several URLs joined by `+` become a fallback transport. Baked into the client bundle at build time, thus readable by every visitor. |
 
 The Worker builds its `connect-src` CSP directive from the same two variables (see `src/index.ts`), so an override is allowed by the CSP without more changes.
+
+For the deployed dashboard, set the production values as Cloudflare Workers Builds variables in the Cloudflare dashboard. Vite gives these priority over the committed `.env` at build time. Thus the paid RPC URL stays out of the repository.
 
 ## Development
 
@@ -35,4 +37,4 @@ pnpm tsc      # typecheck only
 
 ## Deployment
 
-Deployed to Cloudflare with `wrangler` (see `wrangler.jsonc`).
+Deployed to Cloudflare with `wrangler` (see `wrangler.jsonc`). The production environment variables come from the Cloudflare Workers Builds variables (see [Environment Variables](#environment-variables)).
--- a/internal-dashboard/README.md
+++ b/internal-dashboard/README.md
@@ -3,10 +3,26 @@
 Internal-facing dashboard for tracking various operational metrics. Each metric area
 lives on its own tab, mapped to a URL.
 
-- **Curve** (`/curve`) — liquidity on Curve pools _(placeholder for now)_.
+- **DEX** (`/dex`) — liquidity of the tracked Vetro pools on Curve, Sushi and
+  Uniswap. Each pool also has a detail view at `/dex/:poolId`.
+- **Hemi Earn** (`/hemi-earn`) — status of the Hemi Earn agent: keepers,
+  ownership, proxy implementation and its pending vault cooldowns.
+
+`/` and unknown paths redirect to the DEX tab.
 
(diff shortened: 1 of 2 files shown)


internal-dashboard/README.md
```diff
@@ -3,11 +3,27 @@
...
+  ownership, proxy implementation and its pending vault cooldowns.
+
+`/` and unknown paths redirect to the DEX tab.
 
 It's a Vite + React + TypeScript + Tailwind v4 SPA, deployed to Cloudflare via a Worker
 that serves the static assets and injects security headers (see `src/index.ts`).
 
+## Environment Variables
+
+Set these in `internal-dashboard/.env` (or a `.env.local` override) before you run `dev` or `build`.
```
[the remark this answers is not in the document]
gndelia: I think it's pretty obvious

ArturDolzan (dismissed):

packages/cli/src/lib/client.ts
```diff
@@ -9,11 +9,7 @@ const localChainId = 31337;
 const supportedChainIds = [mainnet.id, localChainId];
 
 export async function createVetroClient({ rpcUrl }: GlobalOptions) {
-  const client = createPublicClient({
-    batch: { multicall: true },
-    chain: mainnet,
-    transport: http(rpcUrl),
-  });
+  const client = createMainnetClient(rpcUrl);
```
gabmontes: Question about the CLI's RPC URL contract here: `createMainnetClient` runs the value through `updateRpcUrls`, which splits on `+` to build a fallback transport, while the CLI's `parseRpcUrl` validates a single http(s) URL. Is the CLI meant to accept only one URL, or should it also support several like the dashboard? If single-URL, splitting on `+` seems wrong for an endpoint whose path/key contains a literal `+` (silently truncated vs. the old verbatim `http(rpcUrl)`). If multi-URL is the goal, then splitting is correct — but a `+` inside a single URL would need to be escaped, and that should be documented. Which behavior did you intend?

— Gabi 🤖

pnpm-lock.yaml
```diff
@@ -254,6 +254,9 @@ importers:
       '@types/node':
         specifier: 24.1.0
         version: 24.1.0
```
gabmontes, line 256: Besides linking `@vetro-protocol/core`, this lock update also flips `@typescript-eslint/parser` (6.21.0 → 8.53.1), drops `eslint-import-resolver-typescript`, and re-pins `@types/node` (24.10.13 → 24.1.0). These look unrelated to the PR's change — is this intentional, or leftover drift from an out-of-date branch that should be regenerated from a fresh `master`?

— Auto Gabi

gndelia:
Converting to draft as I need to rebase and review comments

packages/cli/src/lib/client.ts
gndelia: This was rolled back - as it was solved on #735

pnpm-lock.yaml
gndelia: Good catch. Fixed in [0fd51bfc](https://github.com/vetro-protocol/vetro-monorepo/pull/733/commits/0fd51bfc). The lockfile is back to master's version, so this PR now touches only `internal-dashboard/README.md`. `pnpm install --lockfile-only` against master's lockfile makes no change, which confirms the drift was not needed.

gndelia:
PR re-enabled again. Only the README changes are applied.

ArturDolzan (dismissed):

internal-dashboard/README.md
```diff
@@ -3,11 +3,27 @@
...
-- **Curve** (`/curve`) — liquidity on Curve pools _(placeholder for now)_.
+- **DEX** (`/dex`) — liquidity of the tracked Vetro pools on Curve, Sushi and
+  Uniswap. Each pool also has a detail view at `/dex/:poolId`.
+- **Hemi Earn** (`/hemi-earn`) — status of the Hemi Earn agent: keepers,
+  ownership, proxy implementation and its pending vault cooldowns.
+
+`/` and unknown paths redirect to the DEX tab.
 
 It's a Vite + React + TypeScript + Tailwind v4 SPA, deployed to Cloudflare via a Worker
 that serves the static assets and injects security headers (see `src/index.ts`).
```
gabmontes, line 14: This section only covers the local `.env`, but the PR's goal (Closes #732) is that a paid RPC is set as a **Cloudflare Workers Builds variable** — which Vite picks up over the committed `.env` at build time. Since `.env` is tracked in git, a reader following this literally would commit the paid RPC URL into the repo. Worth adding a line here (or in the Deployment section) stating that production values go in the Cloudflare Workers Builds variables and override the committed `.env`.

— Gabi 🤖

gndelia: Fixed in [ee8508c2](https://github.com/vetro-protocol/vetro-monorepo/pull/733/commits/ee8508c26e7602c4939a69de91e49c9877989b45). The Environment Variables section now states that the production values go in the Cloudflare Workers Builds variables, which Vite gives priority over the committed `.env` at build time, and the Deployment section points back to it.

ArturDolzan (approved):

gabmontes (approved):

PR #734, opened 2026-08-17
Seed queries from a static protocol graph

### LINKED ISSUE #724: Make improvements to reduce load time
I've been doing some analysis to improve the performance on the app related to some RPC calls being slow (due to public RPC urls).  
Here I'm laying down some findings which I think with a few changes may improve the page load times.

## Terms used in this issue

- **Protocol graph** — the addresses that connect the contracts: each gateway to its
  pegged token, its treasury, its staking vault, and that treasury's whitelisted tokens.
- **Round trip** — one request to the RPC node, and its answer.
- **Chain of dependent reads** — the app must complete one read before it can start the
  next one, because the first read gives the address for the second.

## Problem

The app reads the protocol graph at run time. Each read gives the address for the next
read. The app cannot start the second read before the first read is complete. However, this approach
is useful if new tokens are whitelisted, or any contract upgrade takes place.

The number of reads is not the problem. wagmi puts all reads of the same tick into one
Multicall3 `eth_call`. Token metadata is also static already, in `knownTokens`. The
problem is the length of the chain of dependent reads.

The Swap page makes these reads for each gateway:

| Step | Read                              | Code                     |
| ---- | --------------------------------- | ------------------------ |
| 1    | `gateway.treasury()`              | `useTreasuryAddress`     |
| 2    | `treasury.whitelistedTokens()`    | `fetchWhitelistedTokens` |
| 3    | `treasury.tokenConfig(token)` × N | `fetchRedeemableTokens`  |

`gateway.PEGGED_TOKEN()` runs in parallel with step 1. Thus the token selector must wait
for **3 sequential round trips**. On a public RPC node this is approximately 0.6 to 1.2
seconds of empty user interface. The app repeats this at each cold load. These values did
not change since the deployment.

The Earn page has a chain of 2 reads. `fetchVaultPeggedToken` reads `vault.asset()`, then
reads `peggedTokensByGateway`.

The API has the same problem. `api/src/analytics.ts`, `api/src/staking-vault.ts` and
`api/src/variable-stake.ts` read the same graph.

## Approach

Get the protocol graph before the user opens a page. Write the result into the React
Query cache. The app then renders immediately, and the usual on-chain queries continue to
run and to correct the cached values (revalidation). The contract stays the source of truth.

There are two options. They are different only in the source of the pre-resolved graph.
The values, the query mechanics and the shared module apply to both options.

### Option A — the API gives the graph

Add `GET /protocol/graph` to `api/`. The endpoint gives the full graph: for each gateway,
its pegged token, its treasury, its staking vault, its whitelisted tokens, and their
configuration. The app makes this one request at start, then calls `setQueryData` for each
related query key.

The API has the necessary parts already. `api/src/analytics.ts`, `staking-vault.ts` and
`variable-stake.ts` read this graph today. `warm-cache.ts` gives cron-warmed KV entries
with a write-through fallback. The new endpoint is a `globalWarmTask` plus a Hono route
with `cache({ cacheControl: "max-age=…" })`. The `/variable-stake/apy` endpoint has the
same structure.

- **Delay:** 3 sequential RPC round trips become 1 HTTP round trip to Cloudflare edge KV.
  The browser can also cache the answer.
- **Old data:** none. A new whitelisted token is visible at the next cron tick. No PR is
  necessary.
- **Risk:** the Swap and Earn pages then need the API. Today only the Analytics page and
  the APY value need it, and they degrade correctly: `useApy` and
  `useCollateralizationRatio` disable themselves if `VITE_VETRO_API_URL` is not set, and
  the user interface shows "-". The graph cannot degrade in this manner. Thus it needs a
  fallback to the on-chain reads. Both code paths must have tests.
- **Also:** the token selector still waits for 1 round trip. Local development and the e2e
  tests get one more component. The e2e tests use a fork of mainnet. If a test changes the
  fork, the answer from the API does not agree with the fork.

### Option B — a static object as `initialData`

Put the graph in a static object in `packages/core`. Give this object to the related
queries as `initialData`. The queries then read the chain at the first render and correct
the values.

- **Delay:** the chain of 3 dependent reads becomes 1 round trip. The token selector
  renders immediately, because the whitelisted tokens and their metadata are available
  without a read. The configuration and the pause switches then arrive together, in one
  batch, and they enable the buttons.
- **Old data:** the object is different from the chain until a PR corrects
  the in-memory object. The generator and the CI check below make this interval short.
  For a new whitelisted token the app corrects itself in the same session.
- **Risk:** the tooling for the static object, and the query mechanics below. A query with
  `staleTime: Infinity` and an initial value never reads the chain again.
- **Advantage:** no new run-time dependency. The app continues to operate if the API is
  down. The API can import the same object and stop its own reads.

### The two options agree

The options are not exclusive. Option B gives the token selector without a read. Option A
can then give the new graph values, in 1 round trip and not 3. If you do both, the static
object becomes a build-time copy of the endpoint.

## Query mechanics

These items are about writes to the cache. They apply primarily to Option B. For Option A
the equivalent is `setQueryData`, which has the same effect on the stale state.

1. **Use `initialData`, and not `placeholderData`.** The fetchers get most of these values
   with `queryClient.ensureQueryData(...)`, and not with `useQuery`. React Query does not
   write `placeholderData` to the cache. Thus `ensureQueryData` continues to wait for the
   network, and the chain of dependent reads stays. `initialData` is cache data, and
   `ensureQueryData` gives it immediately.
2. **`staleTime: Infinity` stops the correction.** `useTreasuryAddress`,
   `useWhitelistedTokens` and `useTokenInfo` set it today. With an initial value and an
   infinite `staleTime`, React Query never reads the chain again. This is the failure that
   this issue must prevent. Give these queries a finite `staleTime` (for example 1 hour)
   and `initialDataUpdatedAt: 0`. The initial value is then stale immediately, so React
   Query reads the chain at mount. A value from the chain stays fresh for the full
   interval.
3. **`ensureQueryData` does not read the chain again.** It gives the cached value and
   stops. Call it with `ensureQueryData({ ...options, revalidateIfStale: true })` to get
   the stale value immediately and a read of the chain in the background.

## The values

Only the graph gets an initial value. The app reads all other values from the chain.

### The contract cannot change these

These values have no setter. Only a proxy upgrade that runs an initializer again can
change them, and `Initializable` prevents this.

| Value                                    | Evidence in the contract               | Code that reads it                                                          |
| ---------------------------------------- | -------------------------------------- | --------------------------------------------------------------------------- |
| `gateway.PEGGED_TOKEN()`                 | set in `Gateway.initialize`, no setter | `fetchPeggedToken`, `usePeggedTokensByGateway`, API                         |
| `stakingVault.asset()`                   | set by `__ERC4626_init`, no setter     | `useVaultAsset`, `fetchStakingVaultForPeggedToken`, `fetchVaultPeggedToken` |
| staking vault to pegged token to gateway | calculated from the two values above   | `useStakingVaultForPeggedToken`, `useShareTokenForPeggedToken`              |
| ERC-20 `name`, `symbol`, `decimals`      | ERC-20 constructor or initializer      | `knownTokens` gives these already. **No work is necessary.**                |

`treasury.PEGGED_TOKEN` is `immutable`, but the app reads the opposite direction. Thus it
is not relevant here.

### The contract can change these, but rarely

Give an initial value, and always read the chain again.

| Value                                                   | Who can change it                                                                                                                          | Effect if the value is old                                                                                                                                                                                                                     |
| ------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `gateway.treasury()`, equal to `peggedToken.treasury()` | `PeggedToken.updateTreasury`, `onlyOwner`. This is a migration operation.                                                                  | The app uses this address for reads only: `whitelistedTokens`, `tokenConfig` and `getPrice`. The gateway finds its treasury internally for mint and redeem. Thus an old address cannot move funds to the wrong contract.                       |
| `treasury.whitelistedTokens()`                          | `addToWhitelist` and `removeFromWhitelist`, `DEFAULT_ADMIN_ROLE`. An admin transfer has a delay of 3 days. `MAX_WHITELISTED_TOKENS` is 10. | A new token is absent until the app reads the chain, then the app corrects itself. A removed token stays in the list until the app reads the chain. The mint then fails with `UnsupportedToken`. The existing simulate code path catches this. |

### The app reads all other values from the chain

This includes `tokenConfig`, the fees, the delays, the cooldowns, the previews, the
balances and the APY. These reads are not expensive after the change. The app knows the
gateway addresses and the whitelisted tokens without a read, so it makes all of them
together, in one batch. `useTokenConfig` therefore stays one query.

**Do not give an initial value to the pause switches later.**
`treasury.tokenConfig(token).depositActive` and `.withdrawActive` are the pause switches
that `DOMAIN.md` describes. `KEEPER_ROLE` can set them at any moment with
`setDepositActive` and `setWithdrawActive`. The app puts them in the conditions that
enable buttons: `deposit.tsx:276` for the Swap button, `oneStepRedeem.tsx:268` for the
redeem button, and `claimRedeemProgressDrawer.tsx:141` for `isPaused`. An old value of
`true` therefore enables a button for an action that the protocol stopped.
`stakingVault.cooldownEnabled()` is the same type of switch.

## Structure

### Location: `packages/core`

There is no circular dependency, because the dependencies point in one direction already.
The repository also has this pattern already:

- `@vetro-protocol/core` is `private: true` and has no dependencies.
- Each other package (gateway, earn, treasury, bridge, morpho-blue-market,
  target-yield-earn) has `"@vetro-protocol/core": "workspace:*"` in **devDependencies**,
  and imports from it. Almost all files in `packages/gateway/src/actions/public/` import
  `isAddressValid`. **No change to a package.json is necessary.**
- `scripts/package-externals.js` keeps this direction. A private workspace can be a
  devDependency only, and esbuild puts it into `_esm`. Thus the published
  `@vetro-protocol/gateway` has no run-time dependency on core.
- Core has addresses already. `packages/core/src/stakingVaultAddresses.ts` is the source.
  `packages/earn/src/stakingVaultAddresses.ts` declares the same values again. It does not
  use `export … from`, because that puts a private package into the published `.d.ts`.
  `knownTokens` imports the same file. Thus the addresses of the pegged tokens and the
  share tokens are in core already.

Only `gatewayAddresses` is different: it is in `packages/gateway`. Three steps correct
this:

1. Move `Gateway`, `gateways` and `gatewayAddresses` to
   `packages/core/src/gatewayAddresses.ts`. Export them from the core index.
2. Write `packages/gateway/src/gatewayAddresses.ts` again, as a declaration of the values
   from core. Use `packages/earn/src/stakingVaultAddresses.ts` as the example: explicit
   types, and the same comment. The public API of `@vetro-protocol/gateway` does not
   change.
3. Add `packages/core/src/protocolGraph.ts`. It imports `./gatewayAddresses.ts` and
   `./stakingVaultAddresses.ts` only. Both files are in core. No package imports in the
   opposite direction.

`gateways[].pegBaseSymbol` is configuration of a gateway, and not an address of a
contract. Put it in the graph entry.

### The module

For Option B this module contains the values. For Option A the same types describe the
body of the `/protocol/graph` answer. Thus the structure is necessary for both options.

One module is sufficient, because only the graph gets an initial value:

```ts
export const protocolGraph = {
  "0xDaD503f8B9d42bb7af3AfC588358D30163e4416F": {   // VUSD gateway
    pegBaseSymbol: "USD",
    peggedToken: "0xCa83DDE9c22254f58e771bE5E157773212AcBAc3",
    stakingVault: sVusdAddress,
    treasury: "0x…",
    whitelistedTokens: ["0xdAC17F…", "0xA0b869…", …],
  },
  "0xCBA2Ffa0AC52d7871a4221a871793Eb788013faB": { … }, // vetBTC gateway
} satisfies Record<Address, GatewayGraph>;
```

The gateway address is the only key. The app operates on mainnet only: `useMainnet`
gives `mainnet`, and `useEthereumClient` uses that chain for all these reads. Add a
level for the chain id if the testnet selector arrives (see the TODO in
`web/src/hooks/useMainnet.ts`).

`web/` and `api/` depend on core already. Thus the API can use the same module and stop
its own reads in `analytics.ts`, `staking-vault.ts` and `variable-stake.ts`.

## Keep the static object correct — Option B only

- Add a generator (a cli command or a script). It reads mainnet and writes the modules
  again. The PR after a change to the whitelisted tokens is then mechanical.
- Add a scheduled CI job. It runs the generator, and it fails or opens a PR if the object
  and the chain do not agree. This makes the interval short.
- Add an e2e assertion: the object must agree with the forked chain. A build with an
  incorrect object then fails before the release.

## Verification

1. Measure the number of round trips at a cold load of the Swap and Earn pages. Use the
   Network panel with throttle. Today the token selector waits for 3 round trips. The
   target is 0 for Option B, and 1 for Option A. Measure also when the buttons become
   active: the configuration and the pause switches then need 1 more round trip.
2. Add a test: a query with an initial value must read the chain at mount. This prevents
   the `staleTime: Infinity` failure in mechanic 2.
3. Add a test: a whitelisted token that is on the chain, but not in the pre-resolved
   graph, must become visible after the app reads the chain. A token that the admin
   removed must become invisible.
4. Option A only: the Swap and Earn pages must render if the API is not available. This
   shows that the on-chain fallback operates.

## Not in this issue

You can change the API to use the shared module in a different PR. The graph work gives a
benefit without this change.

gndelia:
### Description

This PR adds a static object with all the metadata read from the contracts that cannot change, or rarely does. We're seeding the queries from `react-query` with this object using `placeholderData` - this way we get a faster initial rendering, while we revalidate on mount.   
Using `placeholderData` allows us to render and, if there's any change on chain where they differ, the revalidation will cause the correct data to appear. So for example, say a new token is whitelisted for a gateway, then initially all the hardcoded tokens will render, and after revalidation, the new token will appear. This allows us to keep the app looking faster and save some RPC requests, all while allowing graceful revalidation to keep on sync with on-chain data.

Having said all of that, I also added a couple of E2E tests to ensure the graph is kept in sync. New data appearing on-chain won't break the app, but it will break the test, giving us a heads up to update it and keep it in sync.

Note that the issue originally referred to use `initialData` but after testing the implementation, I ended up using `placeholderData` which doesn't fill up the cache, and forces the revalidation I wanted for this feature.

One consequence of that choice: `placeholderData` lives on the observer, so only a query a component subscribes to can render it. The seeded queries are therefore the ones behind a hook: `usePeggedToken`, `usePeggedTokensByGateway`, `useWhitelistedTokens`, `useWhitelistedTokensByGateway`, `useShareTokenForPeggedToken` and `useVaultPeggedToken`. `treasuryAddressOptions` and `vaultAssetOptions` are read only through `ensureQueryData` inside the fetchers, so a seed on them would never render and they keep reading the chain.

### Screenshots

Swap comparison (I recommend checking the videos on slow velocity, or pausing at the appropriate times)

#### Swap page

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/6304c6dc-e257-4118-8146-ea87ffb334c5" controls muted></video> | <video src="https://github.com/user-attachments/assets/e13684a4-05b0-4ed9-9d4b-4ea4919bf0ba" controls muted></video> |

Notice how the initial render of the page used to include a Skeleton loading the tokens, but after this PR it automatically renders the token selector, including the whitelisted tokens

#### Earn page

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/b43e1e44-52ad-4fc7-ac80-55c599d1be0a" controls muted></video> | <video src="https://github.com/user-attachments/assets/b9b1ed35-c27b-4ccb-8dd4-54eff6128225" controls muted></video> |

Notice how the initial page of the render used to show some skeletons, but now renders the rows for the staking vaults

### Related issue(s)

Closes #724

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/core/src/protocolGraph.ts
+++ b/packages/core/src/protocolGraph.ts
@@ -1,0 +1,48 @@
+import type { Address } from "viem";
+
+import type { Gateway } from "./types.ts";
+
+// A hand-maintained copy of the protocol graph on Ethereum mainnet, for
+// faster rendering times skipping initial reading for on-chain values.
+export const gateways: Gateway[] = [
+  {
+    // VUSD gateway
+    address: "0xDaD503f8B9d42bb7af3AfC588358D30163e4416F",
+    pegBaseSymbol: "USD",
+    // VUSD
+    peggedToken: "0xCa83DDE9c22254f58e771bE5E157773212AcBAc3",
+    // sVUSD
+    stakingVault: "0x476310E34D2810f7d79C43A74E4D79405bd7a925",
+    treasury: "0xC8317A10385BE07901A4c9ee3d06E1D83AE378c9",
+    whitelistedTokens: [
+      // USDT
+      "0xdAC17F958D2ee523a2206206994597C13D831ec7",
+      // USDC
+      "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+      // frxUSD
+      "0xCAcd6fd266aF91b8AeD52aCCc382b4e165586E29",
+    ],
+  },
+  {
+    // vetBTC gateway
+    address: "0xCBA2Ffa0AC52d7871a4221a871793Eb788013faB",
+    pegBaseSymbol: "BTC",
+    // vetBTC
+    peggedToken: "0xf196C68233464A16CFDa319a47c21f4cECa62001",
+    // svetBTC
+    stakingVault: "0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e",
+    treasury: "0xd25a7b0b817fD816d0995eC67fb70e75EE65Bd7F",
+    whitelistedTokens: [
+      // WBTC
+      "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
+      // cbBTC
+      "0xcbB7C0000aB88B473b1f5aFd9ef808440eed33Bf",
+      // hemiBTC
+      "0x06ea695B91700071B161A434fED42D1DcbAD9f00",
+    ],
+  },
+];
+
+export const gatewayAddresses: Address[] = gateways.map(
+  (gateway) => gateway.address,
+);
--- a/packages/core/src/protocolGraph.ts
+++ b/packages/core/src/protocolGraph.ts
@@ -1,5 +1,6 @@
 import type { Address } from "viem";
 
+import { sVetBtcAddress, sVusdAddress } from "./stakingVaultAddresses.ts";
 import type { Gateway } from "./types.ts";
 
 // A hand-maintained copy of the protocol graph on Ethereum mainnet, for
@@ -11,8 +12,7 @@ export const gateways: Gateway[] = [
     pegBaseSymbol: "USD",
     // VUSD
     peggedToken: "0xCa83DDE9c22254f58e771bE5E157773212AcBAc3",
-    // sVUSD
-    stakingVault: "0x476310E34D2810f7d79C43A74E4D79405bd7a925",
+    stakingVault: sVusdAddress,
     treasury: "0xC8317A10385BE07901A4c9ee3d06E1D83AE378c9",
     whitelistedTokens: [
       // USDT
@@ -29,8 +29,7 @@ export const gateways: Gateway[] = [
     pegBaseSymbol: "BTC",
     // vetBTC
     peggedToken: "0xf196C68233464A16CFDa319a47c21f4cECa62001",
-    // svetBTC
-    stakingVault: "0x0cB9D84d4bcEc8d3D5B2d99a6F07f4605325987e",
+    stakingVault: sVetBtcAddress,
     treasury: "0xd25a7b0b817fD816d0995eC67fb70e75EE65Bd7F",
     whitelistedTokens: [
       // WBTC
--- a/packages/core/test/protocolGraph.test.ts
+++ b/packages/core/test/protocolGraph.test.ts
@@ -51,16 +51,20 @@ describe("gateways", function () {
     },
   );
 
-  // The app builds `placeholderData` for the token selector out of the graph
-  // plus `knownTokens`. A token missing from `knownTokens` has no metadata, so
-  // the selector falls back to reading the chain and the graph gains nothing.
+  // The app builds `placeholderData` out of the graph plus `knownTokens`: the
+  // token selector from the pegged and whitelisted tokens, the Analytics share
+  // token from the staking vault. A token missing from `knownTokens` has no
+  // metadata, so that query falls back to reading the chain and the graph
+  // gains nothing.
   it.for(gateways)(
     "$address should have metadata for every token",
     function (gateway) {
       expect(
-        [gateway.peggedToken, ...gateway.whitelistedTokens].filter(
-          (address) => !isKnownOnMainnet(address),
-        ),
+        [
+          gateway.peggedToken,
+          ...gateway.whitelistedTokens,
+          ...(gateway.stakingVault ? [gateway.stakingVault] : []),
+        ].filter((address) => !isKnownOnMainnet(address)),
       ).toEqual([]);
     },
   );
--- a/packages/core/test/protocolGraph.test.ts
+++ b/packages/core/test/protocolGraph.test.ts
@@ -1,0 +1,67 @@
+import { type Address, getAddress, isAddressEqual } from "viem";
+import { mainnet } from "viem/chains";
+import { describe, expect, it } from "vitest";
+
+import { gateways } from "../src/protocolGraph.ts";
+import { stakingVaultAddresses } from "../src/stakingVaultAddresses.ts";
+import { knownTokens } from "../src/tokens.ts";
+
+const isKnownOnMainnet = (address: Address) =>
+  knownTokens.some(
+    (token) =>
+      token.chainId === mainnet.id && isAddressEqual(token.address, address),
+  );
+
+describe("gateways", function () {
+  // This test is just to ensure we don't break the default gateway
+  // which is the first one on the list.
+  it("should list the VUSD gateway first", function () {
+    expect(gateways[0].pegBaseSymbol).toBe("USD");
+  });
+
+  it.for(gateways)(
+    "$address should hold checksummed addresses",
+    function (gateway) {
+      const addresses = [
+        gateway.address,
+        gateway.peggedToken,
+        gateway.treasury,
+        ...gateway.whitelistedTokens,
+        ...(gateway.stakingVault ? [gateway.stakingVault] : []),
+      ];
+      expect(addresses).toEqual(
+        addresses.map((address) => getAddress(address)),
+      );
+    },
+  );
+
+  it.for(gateways.filter((gateway) => gateway.stakingVault !== undefined))(
+    "$address should point at a known staking vault",
+    function (gateway) {
+      expect(stakingVaultAddresses).toContain(gateway.stakingVault);
+    },
+  );
+
+  it.for(gateways)(
+    "$address should list each whitelisted token once",
+    function (gateway) {
+      expect(new Set(gateway.whitelistedTokens).size).toBe(
+        gateway.whitelistedTokens.length,
+      );
+    },
+  );
+
+  // The app builds `placeholderData` for the token selector out of the graph
+  // plus `knownTokens`. A token missing from `knownTokens` has no metadata, so
+  // the selector falls back to reading the chain and the graph gains nothing.
+  it.for(gateways)(
+    "$address should have metadata for every token",
+    function (gateway) {
+      expect(
+        [gateway.peggedToken, ...gateway.whitelistedTokens].filter(
+          (address) => !isKnownOnMainnet(address),
+        ),
+      ).toEqual([]);
+    },
+  );
+});
--- a/web/e2e/protocolGraph.spec.ts
+++ b/web/e2e/protocolGraph.spec.ts
@@ -1,0 +1,47 @@
+import { expect, test } from "@playwright/test";
+import { gateways } from "@vetro-protocol/core";
+import { getPeggedToken, getTreasury } from "@vetro-protocol/gateway/actions";
+import { getWhitelistedTokens } from "@vetro-protocol/treasury/actions";
+import { getAddress } from "viem";
+import { asset } from "viem-erc4626/actions";
+
+import { createEthereumClient } from "./anvil";
+
+test.describe("the protocol graph agrees with the chain", function () {
+  for (const gateway of gateways) {
+    test(`gateway ${gateway.address}`, async function () {
+      const client = createEthereumClient();
+
+      const [peggedToken, treasury] = await Promise.all([
+        getPeggedToken(client, { address: gateway.address }),
+        getTreasury(client, { address: gateway.address }),
+      ]);
+
+      expect(
+        getAddress(peggedToken),
+        `Pegged token for gateway ${gateway.address} does not match`,
+      ).toBe(gateway.peggedToken);
+      expect(
+        getAddress(treasury),
+        `Treasury for gateway ${gateway.address} does not match`,
+      ).toBe(gateway.treasury);
+
+      const whitelistedTokens = await getWhitelistedTokens(client, {
+        address: treasury,
+      });
+      expect(whitelistedTokens.map((token) => getAddress(token))).toEqual(
+        gateway.whitelistedTokens,
+      );
+
+      if (gateway.stakingVault) {
+        const vaultAsset = await asset(client, {
+          address: gateway.stakingVault,
+        });
+        expect(
+          getAddress(vaultAsset),
+          `Staking vault ${gateway.stakingVault} for gateway ${gateway.address} don't match their asset and pegged token.`,
+        ).toBe(gateway.peggedToken);
+      }
+    });
+  }
+});
--- a/web/e2e/protocolGraph.spec.ts
+++ b/web/e2e/protocolGraph.spec.ts
@@ -39,7 +39,7 @@ test.describe("the protocol graph agrees with the chain", function () {
         });
         expect(
           getAddress(vaultAsset),
-          `Staking vault ${gateway.stakingVault} for gateway ${gateway.address} don't match their asset and pegged token.`,
+          `Staking vault ${gateway.stakingVault} for gateway ${gateway.address} does not match its asset and pegged token.`,
         ).toBe(gateway.peggedToken);
       }
     });
--- a/web/src/hooks/useTreasuryAddress.ts
+++ b/web/src/hooks/useTreasuryAddress.ts
@@ -1,5 +1,6 @@
 import { queryOptions } from "@tanstack/react-query";
 import { getTreasury } from "@vetro-protocol/gateway/actions";
+import { graphTreasury } from "utils/protocolGraph";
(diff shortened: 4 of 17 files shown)


web/src/hooks/useWhitelistedTokensByGateway.ts
```diff
@@ -20,6 +21,9 @@ export const whitelistedTokensByGatewayOptions = ({
 }) =>
   queryOptions({
     enabled: !!client && !!client.chain && gatewayAddress !== undefined,
+    placeholderData: gatewayAddress
+      ? graphWhitelistedTokens(gatewayAddress)
+      : undefined,
```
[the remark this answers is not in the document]
gndelia: yes.. This is ok and accepted. The approach you mention was the original, but it required further changes and it added more complexity while adding no benefit. `fetchRedeemableTokens ` still requires loading the token config, so RPC calls will be needed regardless to render this.

web/src/hooks/useVaultAsset.ts
```diff
@@ -11,6 +12,7 @@ export const vaultAssetOptions = ({
 }) =>
   queryOptions({
     enabled: !!client,
+    placeholderData: graphVaultAsset(vaultAddress),
```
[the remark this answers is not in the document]
gndelia: It's ok. We want `useVaultAsset` to revalidate while the placeholder renders. This is ok and accepted.

packages/core/test/protocolGraph.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  // The app builds `placeholderData` for the token selector out of the graph
+  // plus `knownTokens`. A token missing from `knownTokens` has no metadata, so
+  // the selector falls back to reading the chain and the graph gains nothing.
+  it.for(gateways)(
+    "$address should have metadata for every token",
+    function (gateway) {
+      expect(
+        [gateway.peggedToken, ...gateway.whitelistedTokens].filter(
+          (address) => !isKnownOnMainnet(address),
+        ),
```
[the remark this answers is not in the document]
gndelia: Fixed in [a2ee899c](https://github.com/vetro-protocol/vetro-monorepo/pull/734/commits/a2ee899c). `gateway.stakingVault` is now included in the metadata assertion when present, so a vault missing from `knownTokens` fails the test instead of silently dropping the share-token placeholder.

web/e2e/protocolGraph.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+        gateway.whitelistedTokens,
+      );
+
+      if (gateway.stakingVault) {
+        const vaultAsset = await asset(client, {
+          address: gateway.stakingVault,
+        });
+        expect(
+          getAddress(vaultAsset),
+          `Staking vault ${gateway.stakingVault} for gateway ${gateway.address} don't match their asset and pegged token.`,
```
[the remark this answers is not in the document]
gndelia: Fixed in [a19371a0](https://github.com/vetro-protocol/vetro-monorepo/pull/734/commits/a19371a0).

web/src/hooks/useWhitelistedTokens.ts
```diff
@@ -19,6 +28,7 @@ const whitelistedTokensOptions = ({
 }) =>
   queryOptions({
     enabled: !!client && !!client.chain,
+    placeholderData,
```
ArturDolzan, line 31: Perhaps it's worth checking what happens to the Swap form's default token when the graph drifts from chain. `getInitialState` runs once on mount, so with the placeholder seeding it the preselected token stays stale even after revalidation replaces the list. Might be worth re-seeding once the real data lands.

web/src/hooks/useTreasuryAddress.ts
```diff
@@ -13,6 +14,7 @@ export const treasuryAddressOptions = ({
 }) =>
   queryOptions({
     enabled: !!client,
+    placeholderData: graphTreasury(gatewayAddress),
```
ArturDolzan: Following up on the earlier `useVaultAsset` thread rather than reopening it. Grepping for `treasuryAddressOptions` and `vaultAssetOptions` turns up only `ensureQueryData` call sites, no `useQuery` anywhere, so those two placeholders may have no observer to render them. Could be worth confirming.

packages/core/src/protocolGraph.ts
```diff
@@ -0,0 +1,48 @@
...
+// faster rendering times skipping initial reading for on-chain values.
+export const gateways: Gateway[] = [
+  {
+    // VUSD gateway
+    address: "0xDaD503f8B9d42bb7af3AfC588358D30163e4416F",
+    pegBaseSymbol: "USD",
+    // VUSD
+    peggedToken: "0xCa83DDE9c22254f58e771bE5E157773212AcBAc3",
+    // sVUSD
+    stakingVault: "0x476310E34D2810f7d79C43A74E4D79405bd7a925",
```
ArturDolzan: Wonder if we could point `stakingVault` at the existing `sVusdAddress` / `sVetBtcAddress` constants instead of repeating the literals. `tokens.ts` already imports them from the same package, so it would keep the two lists impossible to drift.

web/src/hooks/useWhitelistedTokens.ts
gndelia: `getInitialState` ? This is not `next`  baby 😅 

I do not understand what this comment means. 

> What happens to the Swap form's default token when the graph drifts from chain

If on-chain matches the graph, then we're cool. If they don't, the initial render will be "incorrect". How "incorrect" depends on the change: for example, adding a new token is barely noticeable. The default token (I think now it is USDT > VUSD) is rendered, and in the background the missing token is found on-chain and added to the token selector

The user won't notice because the token selector needs to be open for that missing token to be seen as "missing". I think the only scenario that may happen is if USDT is removed from VUSD (very unlikely to happen). 

Additionally, since Vetro exists, only 2 tokens have been added after the Gateways were published (`frxUSD` here and `hemiBTC` on `vetBTC`), so I think the compromise is very good. The key point is that the app DOES NOT stop working if a new token is added without us adding it to the graph - the app reconciles the differences and continues to work. It's up to us to later add the token

web/src/hooks/useTreasuryAddress.ts
gndelia: Confirmed, and fixed in [78c668f3](https://github.com/vetro-protocol/vetro-monorepo/pull/734/commits/78c668f30ba8a2d6c9dcf6d3ecf0c8ea0d43f1e3). Good catch — neither key has a `useQuery` observer, so both placeholders were dead code. Every call site goes through `ensureQueryData` (`fetchWhitelistedTokens`, `useTokenConfig`'s `queryFn`, `fetchVaultPeggedToken`, `fetchStakingVaultForPeggedToken`), and React Query keeps `placeholderData` on the observer without ever writing it to query state. The two `placeholderData` lines are gone, along with the `graphTreasury` and `graphVaultAsset` helpers they were the only callers of. The seeds that do have observers are untouched.

packages/core/src/protocolGraph.ts
gndelia: Fixed in [fefbe7ab](https://github.com/vetro-protocol/vetro-monorepo/pull/734/commits/fefbe7ab4b34fbb99e88b23e5d0e0e63e0bb3d8f). `stakingVault` now points at `sVusdAddress` and `sVetBtcAddress`, and the address comments are gone since the constant names carry the same information. Note this makes the `should point at a known staking vault` test tautological; I left it as is for now.

ArturDolzan (approved):

gabmontes (approved):

PR #738, opened 2026-08-18
Return checksummed addresses from the API

### LINKED ISSUE #736: Ensure API returns checksummed format for addresses when reading data from Subgraphs
### Description

Querying subgraphs returns addresses in lowercase. We should return checksummed formats. See, for example, `getUserExitTickets`. Analyze other subgraphs' queries and return the fields as `checksummed`

### Proposed fix

Use `viem` to convert the addresses to the checksum format after reading them from the subgraph and before returning them as an API response

gndelia:
### Description

The API returned the exit tickets from the subgraph as they came. The subgraph stores addresses in lowercase, so `owner` and `stakingVaultAddress` reached the consumer lowercased, while every other address the API returns is checksummed. A consumer comparing them with a strict equality misses the match — the Earn page groups tickets by `ticket.stakingVaultAddress`, and its optimistic tickets carry the checksummed vault address, so the same vault produced two groups.

`getUserExitTickets` now converts those two fields with `checksumAddress` before the response. I checked every other subgraph query in `api/src`: `getCostBasis` already checksums its vault keys, and the rest (`getExitTicketQueueSize`, the share value, APY and total deposits histories) return no address at all.

The exit ticket response also loses its `receiver` field. `subgraph/schema.graphql` declares it, but no handler in `subgraph/src/earn.ts` ever sets it — `handleWithdrawClaimed` drops the `receiver` the `WithdrawClaimed` event carries — so the API always returned `null` for it. The query no longer asks for it. `cancelTxHash` and `claimTxHash` are nullable for real, so they are now typed `?: Hash | null` to match what the subgraph sends.

Two related changes: `CLAUDE.md` states the convention — worded for any external source, since third-party APIs lowercase addresses too — and `validateMarketId` compares the Morpho loan asset with `isAddressEqual` instead of `===`.

**Commits:**

98b63cb7 Add checksummed address convention for the API
85c92895 Drop the id field exit tickets never return
453fdfbd Return checksummed addresses in exit tickets
194fbe4a Compare the loan asset with isAddressEqual
27a6b132 Allow null in the optional exit ticket fields
027fcb3d Drop the receiver field from exit tickets

### Screenshots

N/A — no UI change.

### Related issue(s)

Closes #736

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -263,6 +263,10 @@ Returns all user's variable stake exit tickets to i.e. allow claiming the withdr
 
 The `owner` and `stakingVaultAddress` fields are returned in the checksummed format.
 
+The `cancelTxHash` and `claimTxHash` fields are optional and mutually exclusive: a
+ticket in cooldown or ready to claim has neither, a cancelled ticket has
+`cancelTxHash` and a claimed ticket has `claimTxHash`.
+
 #### Sample Response
 
 ```jsonc
@@ -275,8 +279,7 @@ The `owner` and `stakingVaultAddress` fields are returned in the checksummed for
     "assets": "1050000000000000000",
     "shares": "1000000000000000000",
     "claimableAt": 1700000000,
-    "cancelTxHash": "0x0000000000000000000000000000000000000000000000000000000000000003", // set once the ticket is cancelled
-    "claimTxHash": "0x0000000000000000000000000000000000000000000000000000000000000004", // set once the ticket is claimed
+    "claimTxHash": "0x0000000000000000000000000000000000000000000000000000000000000004",
   },
 ]
 ```
--- a/api/README.md
+++ b/api/README.md
@@ -261,6 +261,8 @@ Returns a summary of the exit tickets queue for the gateway's staking vault: the
 
 Returns all user's variable stake exit tickets to i.e. allow claiming the withdrawn pegged token.
 
+The `owner`, `receiver` and `stakingVaultAddress` fields are returned in the checksummed format.
+
 #### Sample Response
 
 ```jsonc
--- a/api/README.md
+++ b/api/README.md
@@ -261,7 +261,7 @@ Returns a summary of the exit tickets queue for the gateway's staking vault: the
 
 Returns all user's variable stake exit tickets to i.e. allow claiming the withdrawn pegged token.
 
-The `owner`, `receiver` and `stakingVaultAddress` fields are returned in the checksummed format.
+The `owner` and `stakingVaultAddress` fields are returned in the checksummed format.
 
 #### Sample Response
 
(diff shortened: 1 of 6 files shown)


api/src/variable-stake.ts
```diff
@@ -280,7 +279,14 @@ export async function getUserExitTickets({
       `Got exactly 100 exit tickets for ${address}. Implement pagination!`,
     );
   }
-  return exitTickets;
+  return exitTickets.map((ticket) => ({
+    ...ticket,
+    owner: checksumAddress(ticket.owner),
+    receiver: ticket.receiver
+      ? checksumAddress(ticket.receiver)
+      : ticket.receiver,
```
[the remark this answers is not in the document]
gndelia: Fixed in [27a6b132](https://github.com/vetro-protocol/vetro-monorepo/pull/738/commits/27a6b1320561f6936cb25e3cd0bd0e0791ab5288).

Confirmed against the subgraph: `schema.graphql` declares `receiver`, `cancelTxHash` and `claimTxHash` as nullable `Bytes`, and no mapping ever sets `receiver`. The subgraph returns `null`, not `undefined`, so `ExitTicket` now types all three as `?: T | null`.

ArturDolzan (dismissed):

api/README.md
```diff
@@ -261,20 +261,22 @@ Returns a summary of the exit tickets queue for the gateway's staking vault: the
...
     "requestId": "1",
     "requestTxHash": "0x0000000000000000000000000000000000000000000000000000000000000001",
     "stakingVaultAddress": "0x0000000000000000000000000000000000000002",
     "owner": "0x0000000000000000000000000000000000000001",
     "assets": "1050000000000000000",
     "shares": "1000000000000000000",
     "claimableAt": 1700000000,
-    // Optionally: cancelTxHash, claimTxHash and receiver address
+    "cancelTxHash": "0x0000000000000000000000000000000000000000000000000000000000000003", // set once the ticket is cancelled
+    "claimTxHash": "0x0000000000000000000000000000000000000000000000000000000000000004", // set once the ticket is claimed
```
gabmontes: This sample now shows both `cancelTxHash` and `claimTxHash` present on the same ticket, but they're mutually exclusive: a ticket is either cancelled (`cancelTxHash`) or withdrawn (`claimTxHash`), never both, and both are absent while it's in cooldown/ready. Showing them together implies a state that can't occur, and it also drops the "these fields are optional" signal the previous `// Optionally:` comment carried. Consider showing just one (e.g. `claimTxHash`) with a note that they're optional and mutually exclusive.

— Gabi 🤖

gndelia: Fixed in [64619e81](https://github.com/vetro-protocol/vetro-monorepo/pull/738/commits/64619e81d5e61e85a76eef6900019bdafb7ffd5a). The sample now shows only `claimTxHash`, and a note above it states that `cancelTxHash` and `claimTxHash` are optional and mutually exclusive.

ArturDolzan (approved):

gabmontes (approved):

PR #739, opened 2026-08-18
Add svetBTC to the Bridge page

### LINKED ISSUE #737: Enable SVetBTC on Bridge page

HEre are the OFT addresses https://github.com/vetro-protocol/vetro-contracts/issues/111#issue-4122581591

LEt's enable them on our bridge page

gndelia:
### Description

Enable svetBTC on the Bridge page: the OFT adapter on Ethereum, and the OFT on Hemi, Arbitrum, Base, BSC, and Optimism. The addresses come from vetro-protocol/vetro-contracts#111.

Note: I have not tested with real funds it works

### Screenshots

image

image

image

### Related issue(s)

Closes #737

### Checklist

- [x] Manual testing passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/tokens.ts
+++ b/packages/core/src/tokens.ts
@@ -3,6 +3,14 @@ import { arbitrum, base, bsc, hemi, mainnet, optimism } from "viem/chains";
 import { sVetBtcAddress, sVusdAddress } from "./stakingVaultAddresses.ts";
 import type { Token } from "./types.ts";
 
+// The staking vaults are OZ 5.x upgradeable contracts, which store ERC20 state
+// in an ERC-7201 namespace ("openzeppelin.storage.ERC20") instead of at small
+// sequential slots. Inside that struct, _balances is at offset 0 and
+// _allowances at offset 1, so the slot below is `ERC20StorageLocation + 1`
+// (see `ERC20Upgradeable.sol` in @openzeppelin/contracts-upgradeable v5).
+const erc7201Erc20AllowanceSlot =
+  0x52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace01n;
+
 export const knownTokens: Token[] = [
   {
     address: "0x06ea695B91700071B161A434fED42D1DcbAD9f00",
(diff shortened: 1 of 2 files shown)


ArturDolzan (dismissed):
uACK

gndelia:
E2E are failing to install chromium 🤷🏽

packages/core/src/tokens.ts
```diff
@@ -174,6 +174,68 @@ export const knownTokens: Token[] = [
     chainId: mainnet.id,
     decimals: 18,
     extensions: {
+      allowanceSlot:
+        0x52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace01n,
```
gabmontes: The sVUSD entry has a comment explaining this ERC-7201 slot (`ERC20StorageLocation + 1`); this svetBTC entry uses the same value without it. The slot is correct since it derives from the OZ namespace string, so it's shared across both vaults. Up to you whether it's worth duplicating the comment (or extracting it to a shared const) for the next reader — feel free to skip if you'd rather keep it lean.

— Auto Gabi

packages/core/src/tokens.ts
```diff
@@ -174,6 +174,68 @@ export const knownTokens: Token[] = [
...
+    },
+    logoURI: "https://hemilabs.github.io/token-list/l1Logos/svetbtc.svg",
+    name: "Staked Vetro BTC",
+    symbol: "svetBTC",
+  },
+  {
+    address: "0x54181404A037757eb5271Ee4a02CA51844f25eaA",
+    chainId: arbitrum.id,
+    decimals: 18,
+    extensions: {
```
gabmontes, line 196: `allowanceSlot: 6n` looks copied from the sVUSD OFT layout. It only feeds the gas-estimation state override, so a wrong value just falls back to a real approval tx (no fund risk). Since these weren't tested with real funds, might be worth confirming slot 6 is the actual `_allowances` slot on the svetBTC OFTs — but I'll defer to you on whether that's worth the effort here.

— Gabi 🤖

packages/core/src/tokens.ts
gndelia: Fixed in [b0abd4f9](https://github.com/vetro-protocol/vetro-monorepo/pull/739/commits/b0abd4f9713bc914096f16567002e8c1dacd3621). Extracted the slot into a shared `erc7201Erc20AllowanceSlot` const, so the value is written once and the explanation sits on the declaration.

packages/core/src/tokens.ts
gndelia: Yup, I checked these. They look copied because they're the same 😅  Those that use the `erc7201Erc20AllowanceSlot` is because they are OFT Adapters, but regular OFTs use the slot 6

ArturDolzan (approved):
uACK

gabmontes (approved):

PR #742, opened 2026-08-18
Add the target fixed yield details page header

gndelia:
### Description

Adds the details page for a target fixed yield vault, at `earn/fixed-term/:address`. It shows the breadcrumb and the vault header only — the rest of the page comes in a follow-up PR.

The route sits behind the `fixedTermYield` feature flag.

`VaultHeader` is built for reuse: the variable yield header will use the same component with a different title.

Two supporting changes came out of this work:

- The breadcrumb selector was duplicated between the borrow details page and the new one, so it is now a shared `BreadcrumbSelector` in `components/base/breadcrumb/`. It hides the chevron when there is nothing to switch to — today the target-yield list holds a single vault.
- Both market headers looked like headings through `text-h1` but were `div`s. They are now real `h1`s, with the token logos hidden from the accessibility tree, so the headings read `hemiBTC / VUSD` and `sVUSD / Target fixed yield` instead of repeating the logo alt text.

**Commits:**

74a5e821 Extract a shared BreadcrumbSelector component
9e52a2bb Make the borrow market name a real heading
ecb89d9a Add the target fixed yield details page

### Screenshots

https://github.com/user-attachments/assets/4a39ceaf-f703-4847-8b3f-1bac9afa84e9

### Related issue(s)

Related to #646

### Checklist

- [x] Manual testing passed.


web/src/pages/fixedTermEarnDetails.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const { data: peggedToken } = useVaultPeggedToken(stakingVaultAddress);
+  const { data: shareToken } = useShareToken(stakingVaultAddress);
+
+  if (!peggedToken || !shareToken) {
+    return (
+      <div className="p-8">
+        <Skeleton count={3} height={40} />
+      </div>
+    );
+  }
```
ArturDolzan: Might be worth a quick look at what happens when `useVaultPeggedToken` rejects (RPC hiccup, or the `No gateway found` throw). It looks like we'd sit on the skeleton forever. Perhaps carrying over the `TODO ... issues/146` note from the borrow page would help?

web/src/components/earn/vaultHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  shareTokenSymbol,
+  title,
+}: Props) => (
+  <div className="relative flex h-50 items-center justify-center overflow-hidden border-b border-gray-200 px-4">
+    <img
+      alt=""
+      aria-hidden="true"
+      className="absolute top-1/2 left-1/2 size-70 -translate-x-1/2 -translate-y-1/2 opacity-5 lg:size-100"
+      src={peggedToken.logoURI}
+    />
```
ArturDolzan, line 21: Could be worth confirming the background `img` here. `fetchTokenInfo` falls back to `logoURI: ""` for tokens outside `knownTokens`, and `src=""` makes browsers re-request the page. Probably fine for sVUSD today, but relevant once the real vault list lands.

web/src/components/earn/vaultHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <h1 className="text-h1 relative flex flex-col items-center gap-2 text-center sm:flex-row sm:gap-4">
```
ArturDolzan: Wonder if we can drop `text-h1` here, since `index.css` already applies it to every `h1` in the base layer. The `MarketHeader` in this same PR relies on that, so it'd line the two up.

web/src/components/earn/vaultHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  title: string;
+};
+
+export const VaultHeader = ({
+  peggedToken,
+  shareTokenSymbol,
+  title,
+}: Props) => (
+  <div className="relative flex h-50 items-center justify-center overflow-hidden border-b border-gray-200 px-4">
+    <img
```
gabmontes, line 16: The `text-h1` class here is redundant — the base layer already applies `text-h1` to every `<h1>` (`index.css`). This is also inconsistent with the `MarketHeader` change in this same PR, which drops `text-h1` from its `<h1>` and relies on the base rule. Suggest removing it here too for consistency.

— Gabi 🤖

web/src/components/earn/vaultHeader.tsx
gndelia: All tokens will be known. It would be very weird if we added a staking vault without a known token as a deposit token.

web/src/components/base/breadcrumb/breadcrumbSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+        tabIndex={tabIndex}
+        to={getItemUrl(item)}
+      >
+        {children}
+      </I18nLink>
+    )}
+    renderTrigger={(isOpen, triggerProps) => (
+      <Button {...triggerProps} size="xSmall" variant="tertiary">
+        {trigger}
+        {items.length > 0 && <ChevronIcon direction={isOpen ? "up" : "down"} />}
```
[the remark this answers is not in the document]
gndelia: Done in [b598317](https://github.com/vetro-protocol/vetro-monorepo/pull/742/commits/b598317a848f89e391a6b10c55e210ee55b4b9bf)

web/src/components/earn/vaultHeader.tsx
gndelia: Done in [d04f6bd](https://github.com/vetro-protocol/vetro-monorepo/pull/742/commits/d04f6bd17570610fd00add28cb8d0640d0ebf225)

web/src/components/earn/vaultHeader.tsx
gndelia: See https://github.com/vetro-protocol/vetro-monorepo/pull/742#discussion_r3809664982 (same comment)

web/src/pages/fixedTermEarnDetails.tsx
gndelia: ok so that issue was for the Borrow details page and was never implemented, so I implemented both in [e65e232](https://github.com/vetro-protocol/vetro-monorepo/pull/742/commits/e65e232dd5d374af15620901d91934ca390fd8eb)

ArturDolzan (approved):

gabmontes (approved):

PR #744, opened 2026-08-18
List every icon in a Storybook story

### LINKED ISSUE #727: Add icons to storybook
Storybook should list all icons available in the app. This should help verify whether an icon is already being used or not when we see a Figma design and we don't know if we need to inline it or add it as a new one. Just one story with all the icons in a grid seems enough (and maybe on hover showing the file name - making it easier to find it)

gndelia:
### Description

This way it's easier to explore which icons are already added and which are not. The dark theme story is just to show how they look now, if we ever need to update them for a dark theme

Note: It may not include some icons that are inlined. Ok for now - Let's focus on this list for now.

### Screenshots

https://github.com/user-attachments/assets/42922731-db00-48dc-bd89-0a6894b6c64b

### Related issue(s)

Closes #727

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/stories/icons.stories.tsx
+++ b/web/stories/icons.stories.tsx
@@ -1,0 +1,79 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import type { ComponentType, ReactNode } from "react";
+
+type Icon = {
+  element: ReactNode;
+  name: string;
+  path: string;
+};
+
+const iconComponents = import.meta.glob<Record<string, ComponentType>>(
+  "../src/**/*Icon.tsx",
+  { eager: true },
+);
+
+const iconFiles = import.meta.glob<string>("../src/**/*.svg", {
+  eager: true,
+  import: "default",
+});
+
+const toPath = (globKey: string) => globKey.replace("../", "");
+
+const toFileName = (globKey: string) => globKey.split("/").pop()!;
+
+const icons: Icon[] = Object.entries(iconComponents)
+  .flatMap(([globKey, module]) =>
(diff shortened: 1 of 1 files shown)


gndelia:
E2E failure is unrelated to the story - I'll have to dig on it..

web/stories/icons.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const toFileName = (globKey: string) => globKey.split("/").pop()!;
+
+const icons: Icon[] = Object.entries(iconComponents)
+  .flatMap(([globKey, module]) =>
+    Object.entries(module)
+      .filter(([name]) => name.endsWith("Icon"))
+      .map(([, IconComponent]) => ({
+        element: <IconComponent />,
+        name: toFileName(globKey),
```
gabmontes, line 30: `name` (and the `key`) are derived from the **filename**, not the export name. The `flatMap` over `Object.entries(module)` already supports more than one `*Icon` export per file, but if a file ever exports two icons they'd both get the same `name` — duplicate labels in the grid and a duplicate React `key` (`path + name` collides since `path` also ends in the filename). Today every file has a single icon export so it's harmless, but consider using the actual export name here to make it robust. Minor either way.

— Gabi 🤖

gndelia: Unlikely to happen. The icons folder adds one export per file (which is the icon), so I'm not worried about this

ArturDolzan (approved):

gabmontes (approved):

PR #751, opened 2026-08-20
Add the variable yield details page header

gndelia:
### Description

First part of the variable yield pool details page. It only adds the breadcrumb and the header band — the stats, the charts and the deposit form come in later parts.

The page lives at `earn/variable-yield/:address` and is behind the new `VITE_VARIABLE_YIELD_DETAILS_ENABLED` flag, committed as `false`. It will be set to `true` in Staging.

### Screenshots

image

It uses the same  header component for Fixed term yield

### Related issue(s)

Related to #709

### Checklist

- [x] Manual testing passed.


web/README.md
```diff
@@ -8,17 +8,18 @@ Built with React, Viem, Wagmi, and Tailwind CSS.
...
+| `VITE_PORTAL_API_URL`                 | Yes      | Hemi Portal API base URL (used for token prices).                                                                                 |
+| `VITE_RPC_URL_MAINNET`                | No       | RPC URL(s) for Ethereum mainnet. Several URLs joined by `+` become a fallback transport. Falls back to viem's default when unset. |
+| `VITE_SENTRY_DSN`                     | No       | Sentry DSN. When unset, Sentry is disabled.                                                                                       |
+| `VITE_VARIABLE_YIELD_DETAILS_ENABLED` | No       | Feature flag for the variable yield pool details page. Enabled only when set to `"true"`.                                         |
+| `VITE_WALLET_CONNECT_PROJECT_ID`      | No       | WalletConnect project ID.                                                                                                         |
+| `VITE_VETRO_API_URL`                  | Yes      | Vetro backend API base URL (analytics, APR history, exit tickets, rewards, etc.).                                                 |
```
ArturDolzan: Perhaps it's worth swapping the last two rows while we're already reformatting the table. `VITE_VETRO_API_URL` would come before `VITE_WALLET_CONNECT_PROJECT_ID` alphabetically, which is also the order in `web/.env`.

gndelia: Fixed in [5f04b51e](https://github.com/vetro-protocol/vetro-monorepo/pull/751/commits/5f04b51e). Good catch — the table now matches `web/.env`.

ArturDolzan (approved):

gabmontes (approved):

PR #758, opened 2026-08-21
Fix flaky Earn test

gndelia:
### Description

This PR intends to fix some Earn tests that were failing occasionally.

The problem was the following:

The app redirects automatically from `/` => `/en` => `/en/swap`

The navbar renders on all of these pages (as it's not dependent on the page the user is on- the navbar is always the same). What would happen ocasionally is that the test would start on `/`, and then the test would click on the `/earn` link in the navbar before the last redirect took place. Because of this, the user would be redirected to `/en/earn` and then, on top of that, to `/en/swap` - this caused the Earn page not to render at all - causing the test to fail.

For regular users, this is not a problem because it happens at the milliseconds level. For automated tests, it was "random" as it depended on how long the redirects would take.

Found with AI's help - let's see if this fixes it

The fix: Just start on `/earn` directly at the beginning of the test.

### Screenshots

N/A

### Related issue(s)

Related to #728

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/e2e/earn.spec.ts
+++ b/web/e2e/earn.spec.ts
@@ -89,7 +89,7 @@ test("deposit VUSD into the Earn pool", async function ({
     }),
   ]);
 
-  await page.goto("/");
+  await page.goto("/earn");
 
   // The mock wallet auto-connects silently via EIP-6963 (see fixtures/wallet),
   // so just wait for the header button to switch from "Connect wallet" to the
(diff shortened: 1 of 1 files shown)


ArturDolzan (commented):
Might be worth a quick look at whether these were our only tests clicking a header nav link. Looks like nothing exercises the navbar now. If we want to keep that, starting one test at /en/swap (already locale-prefixed, so no redirect to race) and then clicking Earn would avoid the flake.

gndelia:
> Might be worth a quick look at whether these were our only tests clicking a header nav link. Looks like nothing exercises the navbar now. If we want to keep that, starting one test at /en/swap (already locale-prefixed, so no redirect to race) and then clicking Earn would avoid the flake.

The other tests were on the Swap page (Where this is not a problem), so we're ok.

gabmontes (approved):

ArturDolzan (approved):

PR #760, opened 2026-08-24
Add stats and charts to variable yield details

gndelia:
### Description

The cards already existed on Analytics, so they were moved to `components/`, and both pages render the same one. Note that although it seems like a lot of lines of changes, most of them are actually moving the charts off from `/analytics` and doing slight tweaks so the UI works on both.

Note that on the Earn details page there are 2 spaces: One (on the right) is where the form to deposit/exit will be placed (Future PRs), while in the top section there are 3 stats (APY, staked, on exit) instead of 4 because there are no other value to show yet. We can probably think of another stat to add

### Screenshots

https://github.com/user-attachments/assets/080c5e69-e90b-452e-be9f-c19abf2936ff

### Related issue(s)

Related to #709

### Checklist

- [x] Manual testing passed.


web/src/components/stakedCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+import type { TokenWithGateway } from "types";
+import { formatTokenAmountUsd } from "utils/currency";
+
+type Props = {
+  peggedToken: TokenWithGateway | undefined;
+};
+
+export function StakedCard({ peggedToken }: Props) {
+  const { t } = useTranslation();
+  const { isError: isPeggedTokensError } = usePeggedTokensByGateway();
```
ArturDolzan: Perhaps it's worth double-checking the `usePeggedTokensByGateway()` calls inside these shared cards. On the Earn details page the token comes from `useVaultPeggedToken`, so a failure in that unrelated query would blank all the cards. Wonder if keeping it as a prop (like `TvlCard` still does) is safer.

web/src/pages/variableYieldEarnDetails.tsx
```diff
@@ -94,6 +102,30 @@ const VariableYieldEarnDetailsContent = function ({
...
+            <ShareRatioCard peggedToken={peggedToken} />
+          </div>
+        </div>
+        <div className="bg-gray-100 md:hidden">
+          <StripedDivider />
+        </div>
+        <div className="w-full shrink-0 md:w-[341px] md:border-b md:border-l md:border-gray-200">
+          {/* Placeholder for the deposit/exit form. */}
+          <div className="h-[252px] md:sticky md:top-0" />
+        </div>
```
ArturDolzan: Might be worth a quick look at how this renders under `md`. With `flex-col-reverse` the empty 252px placeholder ends up above the stats, so mobile shows a blank block plus divider before any content. Could we hide it until the form lands?

web/src/pages/variableYieldEarnDetails.tsx
```diff
@@ -94,6 +102,30 @@ const VariableYieldEarnDetailsContent = function ({
...
+          <VariableYieldInfoCards
+            peggedToken={peggedToken}
+            stakingVaultAddress={stakingVaultAddress}
+          />
+          <div className={chartCardContainer}>
+            <ApyHistoryCard peggedToken={peggedToken} />
+          </div>
+          <div className="border-b border-gray-200">
+            <YieldCard peggedToken={peggedToken} />
+          </div>
```
ArturDolzan, line 115: Perhaps it's worth double-checking the horizontal alignment of the Yield allocation card against the APY and share ratio cards around it. `AllocationCard` hardcodes `px-6 md:px-14`, while the neighbours resolve to a different inset at md and lg.

web/src/i18n/locales/es/translation.json
```diff
@@ -387,6 +386,7 @@
         "you-will-stake": "Usted va a hacer stake de",
         "you-will-withdraw": "Usted va a retirar"
       },
+      "staked-label": "{{symbol}} staked",
```
ArturDolzan: Not sure, but maybe the es `staked-label` slipped through untranslated? It matches the en string exactly, while the old analytics key was "Total {{symbol}} depositado" and the file uses "stakeado"/"en stake" elsewhere.

gabmontes: Nit: `"{{symbol}} staked"` leaves "staked" in English. The rest of the es locale always translates it (e.g. `staked-amount` → "Su monto stakeado", `staked-balance` → "Su balance stakeado"), the previous analytics label was "Total {{symbol}} depositado", and the sibling new key `exit-cooldown-label` was translated. Consider `"{{symbol}} en stake"` for consistency.

— Gabi 🤖

web/src/components/stakedCard.tsx
gndelia: Fixed in [6074f9aa](https://github.com/vetro-protocol/vetro-monorepo/pull/760/commits/6074f9aa). The five shared cards now take an optional `peggedTokenError` prop, like `TvlCard` and `CollateralizationCard`. Analytics passes `isPeggedTokensError`; the variable yield details page omits it.

web/src/pages/variableYieldEarnDetails.tsx
gndelia: Fixed in [9a7f8f4e](https://github.com/vetro-protocol/vetro-monorepo/pull/760/commits/9a7f8f4e). The placeholder column is now `max-md:hidden` and the mobile-only `StripedDivider` is removed. Both come back with the deposit/exit form.

web/src/i18n/locales/es/translation.json
gndelia: Fixed in [f8b0c3df](https://github.com/vetro-protocol/vetro-monorepo/pull/760/commits/f8b0c3df). The es `staked-label` is now `"{{symbol}} en stake"`.

gndelia: Fixed in [f8b0c3df](https://github.com/vetro-protocol/vetro-monorepo/pull/760/commits/f8b0c3df). Used `"{{symbol}} en stake"` as suggested.

web/src/pages/variableYieldEarnDetails.tsx
gndelia: It's implemented as designed so we'll leave it like this for now. We can revisit in the future

gabmontes (approved):

ArturDolzan (approved):

PR #762, opened 2026-08-24
Add a swap cooldown read command

gndelia:
### Description

Adds `vetro-cli swap cooldown --gateway <addr>`. It prints the redeem cooldown in seconds, or 0 when the gateway has the withdrawal delay disabled.

Note that the tests duplicate some helpers code - Related to #659 

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.


packages/cli/test/e2e/cooldown.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  it("prints the withdrawal delay when the delay is enabled", async function () {
+    await setWithdrawalDelay({ delay, enabled: true, gateway, rpcUrl });
+
+    const cooldown = await runCli(cooldownOnFork(["--gateway", gateway]));
+
+    expect(cooldown).toBe(delay.toString());
+  });
+
+  it("accepts a non-checksummed gateway address", async function () {
```
gabmontes, line 57: This test doesn't set the withdrawal delay itself — it asserts `cooldown === delay.toString()` relying on the enabled+300 state left by the previous test. If the tests are reordered or this one is run in isolation, it breaks. Consider calling `setWithdrawalDelay({ delay, enabled: true, gateway, rpcUrl })` at the start of this test so it's self-contained. — Gabi 🤖

packages/cli/test/e2e/helpers.ts
```diff
@@ -292,25 +299,105 @@ export const setDepositActive = async function ({
...
+  await impersonateAccount(testClient, { address: admin });
+  await setBalance(testClient, { address: admin, value: parseEther("1") });
+
+  try {
+    const isMaintainer = await readContract(publicClient, {
+      abi: treasuryAbi,
+      address: treasury,
+      args: [maintainerRole, admin],
+      functionName: "hasRole",
+    });
```
gabmontes, line 368: `MAINTAINER_ROLE()` is read from the gateway, but `hasRole`/`grantRole` run against the treasury while `updateWithdrawalDelay`/`setWithdrawalDelayEnabled` run against the gateway. Since `AccessControl` is per-contract, if the gateway enforces the role on its own access control this grant is a no-op and the writes only pass because the admin already holds the role on the gateway. Is the treasury the intended role authority for these gateway functions? If not, the grant/check should target the gateway. — Auto Gabi

gndelia: > if the gateway enforces the role on its own access control 

The Gateway contract doesn't have access control - See [source](https://github.com/vetro-protocol/vetro-contracts/blob/3c063a03940de9aa39538e6a1bce0e8393ccb8e2/src/Gateway.sol#L18)

```solidity
contract Gateway is IGateway, Initializable, ReentrancyGuardTransient {
```

The treasury does implement  `AccessControlDefaultAdminRules` - see [source](https://github.com/vetro-protocol/vetro-contracts/blob/3c063a03940de9aa39538e6a1bce0e8393ccb8e2/src/Treasury.sol#L21)

```solidity
contract Treasury is ReentrancyGuardTransient, AccessControlDefaultAdminRules {
```

So granting and checking can only be done on the Treasury

packages/cli/test/e2e/cooldown.test.ts
gndelia: Fixed in [f9a9e5bb](https://github.com/vetro-protocol/vetro-monorepo/pull/762/commits/f9a9e5bbbd07194ed1f6da69a2a08be5b19b6360).

gabmontes (dismissed):

packages/cli/README.md
```diff
@@ -95,14 +95,15 @@ This is the list of commands available
...
+| Command                                                   | Reads                                             | Returns                                                                                                                                                                 |
+| --------------------------------------------------------- | ------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `vetro-cli swap allowance --token <tok> --account <addr>` | `allowance`                                       | Gateway's spending allowance, in human units. `<tok>` may be a whitelisted token (swap-in) or a pegged token (one-step swap-out; the two-step queue needs no allowance) |
+| `vetro-cli swap cooldown --gateway <addr>`                | `getWithdrawalDelayEnabled`, `getWithdrawalDelay` | Redeem Queue cooldown of the gateway, in seconds. `0` when the delay is disabled — a redeem is then one-step, with no queue.                                            |
```
ArturDolzan: Perhaps it's worth a quick note here that this is the gateway-wide cooldown. `fetchRedeemDelay` in web also factors in `isInstantRedeemWhitelisted`, so a whitelisted account skips the queue even when this prints non-zero. Wondering if a README line (or a later `--account`) would help agents pick the right path.

gndelia: Fixed in [08a42dc3](https://github.com/vetro-protocol/vetro-monorepo/pull/762/commits/08a42dc338e5316e7fa274bb324ee7ae896a92f6). The table now says the command prints the gateway-wide contract value, not a per-account one. I left `--account` out — it changes the command's contract, so it fits better in its own PR.

gabmontes (approved):

ArturDolzan (approved):

PR #763, opened 2026-08-24
Add StakeDAO campaigns to the dashboard DEX page

### LINKED ISSUE #699: Include Merkl campaigns in Dashboard for the DEX page

The DEX page lists pools - we should be able to see the amount of active campaigns in each pool, and the details for each one. Ideally, it should work by autodiscover

gndelia:
### Description

This PR adds StakeDao incentive campaigns. As mentioned before, this is internal tooling, so bear that in mind when considering the quality of the code.

### Screenshots

The campaigns section now uses the appropriate icon depending on the campaign listed.

image

image

image

### Related issue(s)

Closes #699

### Checklist

- [x] Manual testing passed.


internal-dashboard/src/components/dex/campaignsBadge.tsx
```diff
@@ -1,8 +1,50 @@
...
+}: {
+  campaigns: PoolCampaign[];
+  nowSeconds: number;
+}) => (
+  <Tooltip
+    label={<CampaignLabels campaigns={campaigns} nowSeconds={nowSeconds} />}
+  >
+    <CampaignSourceBadge source={campaigns[0].source}>
+      {campaigns.length}
+    </CampaignSourceBadge>
```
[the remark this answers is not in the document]
gndelia: not a problem.

gabmontes (approved):
uACK

internal-dashboard/src/fetchers/fetchPoolCampaigns.ts
```diff
@@ -22,8 +23,10 @@ export const fetchPoolCampaigns = async function (
     return {};
   }
 
-  // TODO add stakeDao campaigns https://github.com/vetro-protocol/vetro-monorepo/issues/699#issuecomment-5357386177
-  const results = await Promise.allSettled([fetchMerklCampaigns(identifiers)]);
+  const results = await Promise.allSettled([
+    fetchMerklCampaigns(identifiers),
+    fetchStakeDaoCampaigns(identifiers),
+  ]);
```
ArturDolzan, line 29: Perhaps it's worth checking what happens when only StakeDAO is down. With two sources now, `allSettled` keeps the partial result and the list shows "No reward campaigns are running on this pool.", so an outage looks the same as no incentives. Maybe worth at least logging the rejected reasons?

gndelia: There's no place to log. I don't think it's worth logging into Sentry for this. The users for this internal dashboard know enough about the domain to, if a campaign is missing, look for the real sources (StakeDao, Merkl) pages to confirm. This is a dashboard - a simplified view so it's more comfortable to interact with all the pieces.

ArturDolzan (approved):

PR #769, opened 2026-08-26
Add brownfi pools

### LINKED ISSUE #748: Add Brownfi pools to the dashboard

These pools need to be added https://app.brownfi.io/clamm/pool

Make sure their rewards are picked up as well (See #699 )

gndelia:
### Description

This PR adds Brownfi pools on Hemi in the dashboard. Because of that, multichain support was added to the table and filters. As mentioned before, notice this is an internal tool so I've focused mostly on funcionality.

### Screenshots

https://github.com/user-attachments/assets/72e33881-17f4-4783-bb6a-907474da64cb

### Related issue(s)

Closes #748

### Checklist

- [x] Manual testing passed.


internal-dashboard/src/lib/brownfiApi.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (body.errors?.length) {
+    const reason = body.errors.map((error) => error.message).join("; ");
+    throw new Error(`BrownFi subgraph error: ${reason}`);
+  }
+  if (!body.data) {
+    throw new Error("BrownFi subgraph returned no data");
+  }
+  return {
+    nativeUsdPrice: Number(body.data.bundles[0]?.maticPriceUSD ?? 0),
+    pools: body.data.pools,
```
gabmontes, line 85: If the subgraph returns an empty `bundles` array, `nativeUsdPrice` silently falls back to `0`. Since every coin's USD price is `derivedMatic * nativeUsdPrice`, the whole pool's TVL would show as `$0` with no error surfaced — the pool list still renders, just with wrong numbers. Consider throwing when `bundles[0]` is undefined so a misconfigured or unsynced subgraph doesn't produce misleading metrics.

— Auto Gabi

internal-dashboard/src/lib/merklApi.ts
```diff
@@ -27,12 +30,18 @@ export const fetchLiveOpportunities = (
   fetch(`${merklProxyApiUrl}/opportunities`, {
     queryString: {
       campaigns: true,
-      chainId: mainnet.id,
+      chainId: trackedChains.map((chain) => chain.id).join(","),
       identifier: identifiers.join(","),
```
gabmontes, line 34: The Merkl v4 API does not support comma-separated `chainId` values — tested live: `chainId=1,43111` returns only Ethereum results. The correct form is repeated params (`chainId=1&chainId=43111`). As written, Hemi Merkl campaigns will never be fetched for BrownFi pools.

Pass an array instead of a joined string so `fetch-plus-plus` serializes it as repeated params:
```ts
chainId: trackedChains.map((chain) => chain.id),
```

— Gabi 🤖

internal-dashboard/src/lib/distribution.ts
```diff
@@ -31,8 +29,8 @@ export const computeDistributions = ({
   tokens.map(function (token) {
     const entries = pools
       .map(function (pool) {
-        const coin = pool.coins.find((candidate) =>
-          isAddressEqual(candidate.address, token.address),
+        const coin = pool.coins.find(
+          (candidate) => candidate.symbol === token.symbol,
```
ArturDolzan: Perhaps it's worth checking the switch to symbol matching here. Since BrownFi pools are discovered from the subgraph and only one leg has to be a tracked token, a Hemi token that just calls itself `VUSD` would get summed into VUSD's total. Wonder if matching against the `knownTokens` addresses per chain would be safer.

internal-dashboard/src/lib/merklApi.ts
gndelia: Incorrect. This works. See url `https://api.merkl.xyz/v4/opportunities?campaigns=true&chainId=1,43111&items=100&status=LIVE` which returns data for both Ethereum and Hemi

image

image

internal-dashboard/src/lib/brownfiApi.ts
gndelia: Fixed in [9209a203](https://github.com/vetro-protocol/vetro-monorepo/pull/769/commits/9209a2037378bc34b0a39cc2e3afdadd35ac5732).

Rather than throw, a coin's `usdPrice` is now `undefined` when nothing could price it, and `tvlUsd` and `baseApy` follow. The pool still lists, but shows an em dash instead of a figure. Throwing would have been invisible anyway — `fetchTrackedPools` wraps each source in `Promise.allSettled`, so BrownFi would have dropped out of the list with no error shown.

This also covers a case beyond the missing bundle: the subgraph reports `derivedMatic: "0"` for any token it can't price, which understated a pool's TVL even with a healthy bundle.

internal-dashboard/src/lib/distribution.ts
gndelia: Good catch — fixed in [f101cec1](https://github.com/vetro-protocol/vetro-monorepo/pull/769/commits/f101cec167317c114d93bbe8b99003102c6b5308).

Symbol matching was needed because tracked tokens are discovered on mainnet, so the same token on Hemi has a different address and `isAddressEqual` never matched. But you're right that it also let an impostor in: a BrownFi pool is listed when only one leg is a tracked token, so the counter leg is arbitrary.

A token now matches its own address plus the `knownTokens` addresses that share its symbol, compared with `isAddressEqual`. `knownTokens` carries VUSD, sVUSD, vetBTC and svetBTC on Hemi, so all four roles resolve. A new test covers a Hemi pool holding a token that only calls itself VUSD.

One trade-off worth knowing: a pegged token from a new gateway that isn't in `knownTokens` yet would match only its mainnet address, so its Hemi pools would be left out until the list is updated.

internal-dashboard/src/index.ts
```diff
@@ -122,14 +115,22 @@ const htmlHeaders = {
...
+
 const graphqlProxies: Record<
   string,
-  { headers?: Record<string, string>; upstream: string }
+  { headers?: Record<string, string>; upstream: (env: Env) => string }
 > = {
-  "/api/sushi": { upstream: "https://production.data-gcp.sushi.com/graphql" },
+  "/api/brownfi": {
+    upstream: (env) =>
+      `https://gateway.thegraph.com/api/${env.THEGRAPH_API_KEY}/subgraphs/id/${brownfiSubgraphId}`,
```
ArturDolzan, line 126: Perhaps it's worth confirming that `THEGRAPH_API_KEY` is scoped to just this subgraph with a query budget in The Graph Studio? `/api/brownfi` is publicly reachable (preview URLs included) and forwards whatever body it gets, so the key's quota is effectively open to anyone who finds the URL.

internal-dashboard/src/lib/poolFilters.test.ts
```diff
@@ -55,16 +76,50 @@ describe("filterPools", function () {
...
   });
 
+  it("matches a Vetro token on every chain it is deployed on", function () {
+    expect(
+      filterPools({
+        campaignPoolIds: [],
+        filters: { ...emptyPoolFilters, trackedSymbols: ["sVUSD"] },
+        pools,
+      }),
+    ).toEqual([svusdUsdt]);
```
ArturDolzan: Might be worth a quick look at this one. `sVUSD` only lives in the mainnet fixture, so the test passes even without cross-chain matching. Filtering on `VUSD` (which is in both `vusdUsdt`/`vusdUsdc` and `hemiVusdUsdc`) would actually pin the behaviour down.

internal-dashboard/src/index.ts
gndelia: No, it's not publicly available - everything is behind CF. See example

image

The same applies to CF Preview. Did you test it to verify it was publicly available? 😄

internal-dashboard/src/lib/poolFilters.test.ts
gndelia: Good catch. Fixed in [f8db2c97](https://github.com/vetro-protocol/vetro-monorepo/pull/769/commits/f8db2c97). The test now filters on `VUSD`, which `vusdUsdt`/`vusdUsdc` and `hemiVusdUsdc` hold under different addresses, so it fails if the filter matches by address.

ArturDolzan (approved):

gabmontes (approved):
uACK

gndelia (commented):
`THEGRAPH_API_KEY` was added to env variables in CF. Merging

PR #774, opened 2026-08-27
Add variable yield details pt4

gndelia:
### Description

This PR adds the "Your position" and "Earned amount" boxes in the Details page of `variable yield`. This is the same as the boxes shown at the top of the Earn page, but filtered down by the pool the user is seeing

### Screenshots

image

image

image

### Related issue(s)

Related to #709

### Checklist

- [x] Manual testing passed.


web/src/utils/queries.ts
```diff
@@ -0,0 +1,9 @@
+import type { UseQueryResult } from "@tanstack/react-query";
+
+export const sumUsdResults = (results: UseQueryResult<number>[]) => ({
+  data: results.every((result) => result.data !== undefined)
+    ? results.reduce((total, result) => total + result.data!, 0)
+    : undefined,
+  isError: results.some((result) => result.isError),
```
gabmontes, line 7: `isError` is computed here but nothing consumes it — `InfoCard` only reads `data`/`isLoading`, and the test `"keeps reporting the error once the other results settle"` confirms a failed vault's stale `data` still gets summed into the total with no visible indicator. The old single-query `fetchTotalStakedUsd`/`fetchEarnedAmountUsd` failed as a whole when any vault failed, keeping the total's snapshot uniformly aged — this refactor changes that. Is silently blending stale per-vault data into the total the intended behavior, or should `isError` get wired into the UI?

Gabi 🤖

web/src/fetchers/fetchEarnedAmountUsd.ts
```diff
@@ -11,68 +11,59 @@ export const fetchEarnedAmountUsd = async function ({
...
-      const earnedAssets = userStakedAssets - costBasis;
-      return tokenAmountToUsd({
-        amount: earnedAssets,
-        prices,
-        token: peggedToken,
-      });
-    }),
-  );
+  const costBasis = costBases[stakingVaultAddress] ?? 0n;
+  if (costBasis === 0n) {
```
gabmontes, line 60: This treats a cost basis of exactly `0n` the same as "not synced yet," short-circuiting to $0 earned even for a position with a legitimately zero cost basis. Pre-existing logic (carried over from the old multi-vault map), but this PR gives it more visibility via the new single-vault `EarnedAmountCard`. Flagging since it can now show a misleading $0.00 for a real position.

Auto Gabi

web/test/fetchers/fetchEarnedAmountUsd.test.ts
```diff
@@ -117,22 +117,14 @@ describe("fetchEarnedAmountUsd", function () {
...
       queryClient,
-      stakingVaultAddresses: [vault1],
+      stakingVaultAddress: vault1,
     });
 
     expect(result).toBe(-2);
   });
 
-  it("sums P&L across multiple vaults", async function () {
+  it("prices the P&L with the vault's own pegged token", async function () {
```
gabmontes, line 126: The old "sums P&L across multiple vaults" test also covered `costBases` being indexed correctly by `stakingVaultAddress` when the cache holds entries for other vaults too. The updated tests only populate one vault's entry, so that indexing guarantee isn't covered anymore, even though `costBasisQueryOptions` still caches all vaults together in production. Could we add back a case with multiple vault entries in `costBases`?

Gabi 🤖

web/src/components/bridgeForm/tokenChainLogo.tsx
```diff
@@ -4,8 +4,10 @@ import { TokenLogo } from "components/tokenLogo";
 import { getChainById } from "networks";
 import type { ComponentProps } from "react";
 
+export type TokenChainLogoSize = ComponentProps<typeof ChainLogo>["size"];
```
gabmontes, line 7: `TokenChainLogoSize` is now typed from `ChainLogo`'s size union instead of `TokenLogo`'s, dropping the `xSmall` option `TokenLogo`/`LogoSize` just gained in this PR. Nothing currently passes `xSmall` through `TokenChainLogo`, so this doesn't break today — was narrowing it to `ChainLogo`'s size intentional, or should `xSmall` stay available here too?

Auto Gabi

gabmontes (commented):
A few questions/notes surfaced during review — nothing blocking, mostly checking intent.

web/src/utils/queries.ts
gndelia: It's added for completeness (It's odd to calculate data and isLoading but not isError), but `InfoCard ` handles errors, so it works ok (if no data and not loading, the only option is an error).

A disconnected wallet renders fine - see

image

Let me know if you want it removed but otherwise I'd leave it.

web/src/fetchers/fetchEarnedAmountUsd.ts
gndelia: I'm not following. Returning `0n` is not taken as "loading" or "not synced yet".

- If the user has no shares, then earned is 0. I don't think we should be showing "past" earnings, imho.
- If the subgraph is not synced, the api fails (or should)
- IF the subgraph is syncing, I don't think there isn't much to do 😅 

I'd keep it like this, open to revisit (although in a new issue, since this is existing behaviour)

web/test/fetchers/fetchEarnedAmountUsd.test.ts
gndelia: `fetchEarnedAmountUsd` used to check the earned amount for multiple vaults; now it does so only for one. Because of that, it makes no sense to consider multiple vaults in the tests.

Now, multiple tests are handled by using `useQueries` (from `react-query`), plus `sumUsdResults`, which has its own suite of unit tests added. So I think there's nothing to do here.

web/src/components/bridgeForm/tokenChainLogo.tsx
gndelia: nope, this narrowing is ok.

ArturDolzan (approved):

gabmontes (approved):

PR #779, opened 2026-08-28
Link the Earn pool bar to the pool details page

gndelia:
### Description

This PR adds the navigational link from the earn page to the details one. It also updates the E2E tests and enables the feature flag.

In the next PR, I'll be dropping all the old code that now will become unreachable if the flag is off - and I will remove it too.

### Screenshots

https://github.com/user-attachments/assets/db7e7346-ac52-4bb7-a311-0eabc400b93a

### Related issue(s)

Related to #709

### Checklist

- [x] Manual testing passed.


gabmontes (dismissed):

web/e2e/earn.spec.ts
```diff
@@ -178,31 +198,11 @@ test("deposit VUSD into the Earn pool", async function ({
...
-
-  const fromPoolsBadge = page.getByText(/from 1 pool/i);
-  await expect(fromPoolsBadge).toBeVisible({ timeout: 20_000 });
-
-  await fromPoolsBadge.hover();
-  // The tooltip lists the pool's staked balance as "(<amount> VUSD)".
-  await expect(page.getByText(/\([\d.,]+\s*VUSD\)/)).toBeVisible({
-    timeout: 10_000,
-  });
+  await expectPositionShares({ page, shares: svusdAfter });
```
ArturDolzan, line 205: Not sure, but maybe we're losing the only e2e coverage of the "From 1 pool" badge and its tooltip here? That stat still lives on the Earn page and isn't part of the code being removed next, so it might be worth keeping the check.

gndelia: Good catch — you're right. `FromPoolsBadge` still renders on the Earn page and isn't gated by the flag, so it survives the cleanup PR, and no other e2e covered it.

Fixed in [38cb2d75](https://github.com/vetro-protocol/vetro-monorepo/pull/779/commits/38cb2d758f4222fb375da4a71e27034e711462f7). The deposit test now goes back to `/earn` after asserting the position card and re-checks the badge plus its `(<amount> VUSD)` tooltip.

ArturDolzan (approved):

gabmontes (approved):

PR #781, opened 2026-08-28
Add a swap send-to-queue command to the CLI

gndelia:
### Description

Adds `vetro-cli swap send-to-queue`, step 1 of the two-step redeem. It prints the `requestRedeem` calldata for a pegged token. The gateway comes from `--from`. There is no receiver and no output token, because `requestRedeem` has neither — the token to redeem into is chosen at step 2.

The command fails early when the gateway has the queue disabled, or when the amount rounds down to zero, instead of emitting calldata that reverts.

Also corrects the README: the two-step queue does need an allowance, because `requestRedeem` moves the pegged token with `safeTransferFrom`.

### Screenshots

N/A — CLI only.

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.


packages/cli/src/features/swap/commands/sendToQueue.ts
```diff
@@ -0,0 +1,61 @@
...
+      });
+
+      const delayEnabled = await getWithdrawalDelayEnabled(client, {
+        address: peggedToken.gatewayAddress,
+      });
+      if (!delayEnabled) {
+        throw new Error(
+          `The redeem queue is disabled: redeem "${options.from}" in one step instead`,
+        );
+      }
```
ArturDolzan, line 43: Could be worth confirming the queue gate here. `fetchRedeemDelay` and `redeem.ts` both pair `getWithdrawalDelayEnabled` with `isInstantRedeemWhitelisted`, and DOMAIN.md lists both conditions. Wondering if a whitelisted agent would get calldata that reverts.

gndelia: Tricky. Technically speaking, the contract does not revert (See code [here](https://github.com/vetro-protocol/vetro-contracts/blob/3c063a03940de9aa39538e6a1bce0e8393ccb8e2/src/Gateway.sol#L265)) so even if you're whitelisted, you can send it to the queue if you want. The operation is not prohibited. 

What would I do here? Throw because the user could go the other path?. I think I'd leave it as here. On the UI is different because based on that check I could show different UI flows (which we do)

gndelia: I'm saying that ☝🏽  to say that, for now I'm leaving it like this

ArturDolzan (approved):

packages/cli/src/features/swap/commands/sendToQueue.ts
```diff
@@ -0,0 +1,61 @@
...
+  swap
+    .command("send-to-queue")
+    .description(
+      "Print the calldata to send a pegged token to the redeem queue",
+    )
+    .requiredOption(
+      "--from <token>",
+      "Pegged token to send to the queue, by symbol or address",
+    )
+    .requiredOption("--amount <n>", "Amount in human units", parseAmount)
```
gabmontes, line 23: > "in human units"

😆

gabmontes (approved):

packages/cli/src/features/swap/commands/sendToQueue.ts
gndelia: 😅

Issue #772, opened 2026-08-27
Stake withdraw forms can switch instant/queued mid-transaction if `canInstantWithdraw` refetches


gabmontes:
Both `StakeWithdrawForm`s (`stakeDrawer/` and its `stakeForm/` copy) read `useCanInstantWithdraw` live and use it directly to pick the mutation and render progress steps — it's never snapshotted at submit time.

**Scenario:** submit a queued withdrawal while `canInstantWithdraw` is `false` → wallet-signing blurs the tab → `QueryClient` has no `defaultOptions` (`web3Provider.tsx`), so default refetch-on-focus applies → value flips to `true` on refocus → `withdrawMutation` and the progress drawer now render the 1-step "instant" UI while the in-flight tx is still the 2-step queued withdrawal (funds actually in cooldown). Reverse flip causes the same mismatch.

Low likelihood (requires eligibility to change mid-flow). Fix is cheap in `stakeForm/`, which already snapshots `{amount, assets}` on submit — add `canInstantWithdraw` to that snapshot. `stakeDrawer/` has no snapshot state yet, needs one added.

Trynax:
Hey @gabmontes @gndelia  I’m not sure if you accept external contributions, but I’d love to contribute here. I traced this issue and can put together a focused fix by snapshotting canInstantWithdraw at submit time in both withdrawal forms, Does that approach match the intended fix?

gndelia:
@Trynax feel free to send a PR, and we will review it whenever we can.

gndelia:
After carefully reviewing this issue, I think we should take a better approach. As #782 shows, this can escalate easily with copying variables and keeping internal states around that are not really worth it.

For 99.99% of the users, `canInstantWithdraw` will be `false`. This flag is created for whitelisting certain internal wallets for a variety of internal OPs. Most of these wallets, if any, will probably be automated and won't use the UI either (although the UI does support it, just in case)

So I think the scenario this issue presents is extremely unlikely. If anything, I think we should just make `canInstantWithdraw` not refetch once loaded and call it a day. That would be a simple change in the hook, and we can skip most of the problems in keeping variables copied around. If this ever becomes a problem, let them come back as a bug.

Trynax:
> After carefully reviewing this issue, I think we should take a better approach. As [#782](https://github.com/vetro-protocol/vetro-monorepo/pull/782) shows, this can escalate easily with copying variables and keeping internal states around that are not really worth it.
> 
[...]

PR #782, opened 2026-08-30
Preserve withdrawal path after submission

Trynax:
### Description

Snapshot the instant or queued withdrawal path when the user submits a withdrawal. This prevents `canInstantWithdraw` refetches from changing the flow while the transaction is pending.

### Screenshots

N/A 

### Related issue(s)

Related to #772

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/web/src/pages/earn/components/stakeForm/stakeWithdrawForm.tsx
+++ b/web/src/pages/earn/components/stakeForm/stakeWithdrawForm.tsx
@@ -110,6 +110,7 @@ export function StakeWithdrawForm({
   const [submitted, setSubmitted] = useState<{
     amount: string;
     assets: bigint;
+    canInstantWithdraw: boolean | undefined;
   } | null>(null);
 
   const instantTracking = useActivityTracking({
@@ -229,7 +230,11 @@ export function StakeWithdrawForm({
   function handleSubmit(e: FormEvent) {
     e.preventDefault();
     if (!inputError) {
-      setSubmitted({ amount: inputValue, assets: amountBigInt });
+      setSubmitted({
+        amount: inputValue,
+        assets: amountBigInt,
+        canInstantWithdraw,
+      });
       onResetSteps();
       onDrawerOpenChange(true);
       withdrawMutation.mutate();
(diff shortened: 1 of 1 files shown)


gndelia:
@Trynax There are conflicts in your PR - please rebase and solve them

web/e2e/earn.spec.ts
```diff
@@ -570,6 +586,190 @@ test("withdraw VUSD from the Earn pool (two-step cooldown exit)", async function
   });
 });
 
+test("keeps the submitted withdrawal path when instant eligibility changes", async function ({
```
gndelia: I think this scenario is very specific and unlikely to happen, so I wouldn't add a full E2E test to it. Additionally, there's no observable behaviour for the user

Trynax:
> @Trynax There are conflicts in your PR - please rebase and solve them

Thanks @gndelia  removed the test and resolved the conflicts.

gndelia:
@Trynax sorry - there are conflicts again 😅

Trynax:
Fixed again @gndelia 😅

gabmontes (commented):
The changes look fine but it does not seem to be a complete fix for #772. Let's modify the PR description. Merging this one should **not** automatically close that issue.

Trynax:
> The changes look fine but it does not seem to be a complete fix for #772. Let's modify the PR description. Merging this one should **not** automatically close that issue.

I’ve updated the pr description and changed the issue reference to “Related to #772”.

gndelia:
LEt's update the PR description as we're no longer adding an E2E test

Trynax:
> LEt's update the PR description as we're no longer adding an E2E test

Thanks i have updated it.

web/src/pages/earn/components/stakeForm/stakeWithdrawForm.tsx
```diff
@@ -229,7 +230,11 @@ export function StakeWithdrawForm({
...
     if (!inputError) {
-      setSubmitted({ amount: inputValue, assets: amountBigInt });
+      setSubmitted({
+        amount: inputValue,
+        assets: amountBigInt,
+        canInstantWithdraw,
+      });
       onResetSteps();
       onDrawerOpenChange(true);
       withdrawMutation.mutate();
```
gabmontes, line 240: `withdrawMutation` and `tracking` are still derived from the live `canInstantWithdraw`, not the new `submitted.canInstantWithdraw` snapshot. If `canInstantWithdraw` flips after this `.mutate()` call, `withdrawMutation` starts pointing at the other, idle `useMutation` instance, so `isPending` reads `false` while the real tx is still in flight. Since the drawer can be dismissed anytime (click-outside/Escape, no pending guard in `useOverlay`), that re-enables the submit button and lets the user fire a second concurrent withdrawal. Same root cause as #772, just the other two spots it mentions.

— Gabi 🤖

gndelia:
@Trynax After reviewing the issue and the original problem, I think we should change the approach. I'm sorry for making you waste your time 😅. See https://github.com/vetro-protocol/vetro-monorepo/issues/772#issuecomment-5586542529 for further details.

Feel free to send a new PR with the proposed approach. It should be far simpler than this one. Based on that, I'm closing this one.

PR #787, opened 2026-08-31
Fix skeleton alignment while prices load

### LINKED ISSUE #778: Skeleton is not properly aligned while loading prices on Earn details drawer
### Description

When staking on the Earn (including the changes on https://github.com/vetro-protocol/vetro-monorepo/issues/709), the skeleton that load prices are not aligned. Most of the time this isnt' a problem because by the time this drawer has opened, the prices were loaded already - we should fix it nonetheless

Image

Notice that this may fixed after https://github.com/vetro-protocol/vetro-monorepo/issues/765 is implemented / unified

gndelia:
### Description

Fixes some alignment issues in the loading skeletons and adds the missing 4px margin after the `$`.

### Screenshots

image

(Example from storybook)

### Related issue(s)

Closes #778

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/components/base/fiatValue.tsx
+++ b/web/src/components/base/fiatValue.tsx
@@ -44,7 +44,7 @@ const RenderFiatValueUnsafe = function ({
   }
 
   // Loading state (either balance or prices are loading)
-  return <Skeleton className="h-full" inline width={32} />;
+  return <Skeleton inline width={32} />;
 };
 
 export const RenderFiatValue = (
--- a/web/src/components/base/fiatValue.tsx
+++ b/web/src/components/base/fiatValue.tsx
@@ -44,7 +44,7 @@ const RenderFiatValueUnsafe = function ({
   }
 
   // Loading state (either balance or prices are loading)
-  return <Skeleton className="h-full" containerClassName="w-8" />;
+  return <Skeleton className="h-full" inline width={32} />;
 };
 
 export const RenderFiatValue = (
--- a/web/src/components/swapForm/swapProgressDrawer.tsx
+++ b/web/src/components/swapForm/swapProgressDrawer.tsx
@@ -64,7 +64,7 @@ export function SwapProgressDrawer({
             amount={fromAmount}
             detail={
               <>
-                $
+                <span className="mr-1">$</span>
                 <RenderFiatValue
                   token={fromToken}
                   value={parseTokenUnits(fromAmount, fromToken)}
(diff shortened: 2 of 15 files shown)


web/src/components/swapForm/swapProgressDrawer.tsx
```diff
@@ -64,7 +64,7 @@ export function SwapProgressDrawer({
             amount={fromAmount}
             detail={
               <>
-                $
+                <span className="mr-1">$</span>
```
ArturDolzan: Wonder if we should pull `<span className="mr-1">$</span>` into a small shared `DollarSign` component. With this PR it lands in 13 spots, plus two local `DollarSign` helpers in `swapFees`/`bridgeFees` doing the same thing via `gap-x-1`.

web/src/components/base/fiatValue.tsx
```diff
@@ -44,7 +44,7 @@ const RenderFiatValueUnsafe = function ({
   }
 
   // Loading state (either balance or prices are loading)
-  return <Skeleton className="h-full" containerClassName="w-8" />;
+  return <Skeleton className="h-full" inline width={32} />;
```
ArturDolzan: Not sure, but maybe `h-full` is a no-op here now. The skeleton's wrapper span has no definite height, so `height: 100%` should resolve to auto. Could be worth a quick Storybook check before dropping it.

web/src/components/swapForm/swapProgressDrawer.tsx
gndelia: Good call — the counts were exactly right. Fixed in [8b9ab34c](https://github.com/vetro-protocol/vetro-monorepo/pull/787/commits/8b9ab34c6c932dd31bda7fefbef8687d877ef9d9).

`DollarSign` now lives in `components/base` with stories, and replaces all 13 spots plus the two local helpers. The fees helpers needed their `Container` gap dropped too, otherwise `gap-x-1` and `mr-1` would have stacked to 8px — verified in Storybook that they still render at 4px.

web/src/components/base/fiatValue.tsx
gndelia: You were right. Fixed in [bbe9191d](https://github.com/vetro-protocol/vetro-monorepo/pull/787/commits/bbe9191d48e6e1130848765710127e1fa6760114), dropped from `cryptoValue` too.

Checked it in Storybook — with and without `h-full` the measurements are identical: the fiat skeleton is 32×13 in a 20px paragraph, the crypto one 80×36 in a 40px row. It was already dead before this PR.

gabmontes (approved):

ArturDolzan (approved):
LGTM.

PR #788, opened 2026-08-31
Add a swap is-instant-redeem read command

gndelia:
### Description

Adds `vetro-cli swap is-instant-redeem --account <addr> --gateway <addr>`. It wraps `isInstantRedeemWhitelisted` and prints whether an address is whitelisted to skip a gateway's redeem queue.

`--gateway` is required. The issue lists only `--account`, but the read is per-gateway and there is no token to infer one from, so the gateway is explicit here, like on the other gateway-level reads.

### Screenshots

N/A — CLI only.

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.


packages/cli/test/e2e/helpers.ts
```diff
@@ -377,32 +374,49 @@ export const setWithdrawalDelay = async function ({
...
-          account: admin,
-          address: treasury,
-          args: [maintainerRole, admin],
-          functionName: "grantRole",
-        }),
-      });
-    }
+  }
+  return admin;
+};
```
gabmontes, line 401: `impersonateMaintainer` starts impersonation and can throw after (the `grantRole` write may revert) with no try/finally of its own. Callers `await` it before opening their own try block, so on that throw `stopImpersonatingAccount` never runs and the account stays impersonated for the rest of the fork's lifetime — the exact case the try/finally in `setDepositActive`/elsewhere guards against. Pre-refactor this was covered since the whole sequence lived in one try/finally.

Worth giving `impersonateMaintainer` its own cleanup on failure.

— Gabi bot

ArturDolzan: Following up on Gabi's thread here rather than opening a new one. Perhaps it's worth confirming that a `grantRole` revert can't leave the admin impersonated now. Both callers await this before their own `try`, so the old `finally` no longer covers it.

gndelia: > Perhaps it's worth confirming that a grantRole revert can't leave the admin impersonated now

I'll fix it, but do note that all of this is running on a local fork 😅

gndelia: Fixed in [6dc3da2c](https://github.com/vetro-protocol/vetro-monorepo/commit/6dc3da2c). `impersonateMaintainer` now wraps the role check and grant in a try/catch that calls `stopImpersonatingAccount` and rethrows, so a `grantRole` revert no longer leaves the admin impersonated.

gabmontes (approved):

ArturDolzan (approved):

PR #794, opened 2026-09-01
Stop price network calls in Storybook

### LINKED ISSUE #731: Avoid Pricing calls in FiatValue stories
### Description

Components/FiatValue in Storybook makes real network calls. The story seeds the price cache, so the first render is clean. After 60 seconds React Query refetches, and the story calls the portal price API and the mainnet RPC. The calls repeat every 60 seconds while the story stays open.

  Cause:

  - RenderFiatValue (web/src/components/base/fiatValue.tsx:24) calls usePrices internally.
  - pricesOptions (web/src/hooks/usePrices.ts:29) sets refetchInterval: 60_000 inline. Inline query options beat queryClient.setQueryDefaults and the QueryClient defaultOptions, so a story file cannot stop the refetch.
  - fetchPrices (web/src/fetchers/fetchPrices.ts:56) calls the portal API directly, not through the viem client. A stub viem transport stops only the RPC POSTs.
  - fetchPrices does not catch errors. The portal API blocks localhost, so the refetch usually rejects in local Storybook. pricesStatus becomes error, and fiatValue.tsx:41 returns `-`.

Result: the Loading and Error stories show `-` instead of the skeleton after about one minute. The WithValue story keeps its value, because React Query holds the seeded data.

`web/stories/fiatValue.stories.tsx` is the only story with a QueryClientProvider today, so it is the only story affected. Every new story that mounts a hook-driven component gets the same problem.

###  Steps to reproduce
  
  1. Run pnpm --filter web storybook.
  2. Open Components/FiatValue → Loading.
  3. Open the network tab and wait 60 seconds.
  4. See the POSTs to the RPC and the GETs to the price API. The skeleton changes to -.

###  Environment

  - OS: macOS
  - Runtime/Browser/Version: Storybook 10.4.6, @storybook/react-vite, Node 22
  - Wallet: not applicable

###  Proposed fix
  
  Four options, from smallest to cleanest:

  1. sb.mock in the preview file. Add sb.mock(import("../src/hooks/usePrices")) to `web/.storybook/preview.tsx`. Storybook 10.4.6 supports it. Limit: mocks must be declared in the preview file, so the mock is global to every story. Each story that needs a price must then set the mock return value itself.
  2. `msw-storybook-addon`. Intercept the HTTP calls at the service-worker level. It stops the portal GETs and the RPC POSTs together, and it works for any future story. Cost: one more dev dependency and a handler set to maintain.
  3. Make RenderFiatValue take the price as a prop. The component becomes pure, the story needs no provider, and no mock is necessary. This is the cleanest option, but it changes production code and every call site.
  4. Move refetchInterval out of pricesOptions and into the useQuery call in usePrices, so pricesOptions stays a plain fetch descriptor. A story that renders through the hook still refetches, so this option alone does not fix the story — it only makes option 1 unnecessary for fetcher call sites.

gndelia:
### Description

The FiatValue stories rendered a real price query. React Query refetched every 60 seconds, so the stories hit the portal price API and the mainnet RPC on a loop. The portal API blocks localhost, so the refetch failed and the Loading and Error stories showed `-` instead of the skeleton.

`fetchers/fetchPrices` is now automocked in the Storybook preview file, and the story supplies the resolved value. Verified in the browser over 80 seconds: without the mock the Loading story makes 3 calls to `/prices` and 4 to the mainnet RPC; with it, zero.

### Screenshots

N/A — Storybook only, no product UI change.

### Related issue(s)

Closes #731

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/.storybook/preview.tsx
+++ b/web/.storybook/preview.tsx
@@ -1,8 +1,11 @@
 import type { Preview } from "@storybook/react";
+import { sb } from "storybook/test";
 
 import "react-loading-skeleton/dist/skeleton.css";
 
 import "../src/index.css";
+
+sb.mock(import("../src/fetchers/fetchPrices.ts"));
 
 const preview: Preview = {
   parameters: {
--- a/web/stories/fiatValue.stories.tsx
+++ b/web/stories/fiatValue.stories.tsx
@@ -1,10 +1,12 @@
 import type { Meta, StoryObj } from "@storybook/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import type { Token } from "@vetro-protocol/core";
+import { mocked } from "storybook/test";
 import { parseUnits } from "viem";
 import { createConfig, http, WagmiProvider } from "wagmi";
 
 import { RenderFiatValue } from "../src/components/base/fiatValue";
+import { fetchPrices } from "../src/fetchers/fetchPrices";
 import { mainnet } from "../src/networks/mainnet";
 
 const usdc: Token = {
@@ -15,6 +17,8 @@ const usdc: Token = {
   name: "USD Coin",
   symbol: "USDC",
 };
+
+const prices = { USDC: "1.00" };
 
 const wagmiConfig = createConfig({
   chains: [mainnet],
@@ -28,9 +32,13 @@ const queryClient = new QueryClient({
     },
   },
 });
-queryClient.setQueryData(["prices", mainnet.id], { USDC: "1.00" });
+queryClient.setQueryData(["prices", mainnet.id], prices);
 
 const meta = {
+  beforeEach() {
+    mocked(fetchPrices).mockResolvedValue(prices);
+    return () => mocked(fetchPrices).mockReset();
+  },
   component: RenderFiatValue,
   decorators: [
     (Story) => (


ArturDolzan (approved):

web/.storybook/preview.tsx
```diff
@@ -1,9 +1,12 @@
 import type { Preview } from "@storybook/react";
+import { sb } from "storybook/test";
 
 import "react-loading-skeleton/dist/skeleton.css";
 
 import "../src/index.css";
 
+sb.mock(import("../src/fetchers/fetchPrices.ts"));
```
gabmontes, line 8: Non-blocking: `sb.mock` here is project-wide and static (no per-story scoping, no `sb.unmock()` — confirmed in Storybook's docs), so it now covers every story, not just FiatValue's. Right now no other story renders a component that calls `usePrices`/`fetchPrices`, but if one does in the future without its own `mockResolvedValue`, that component will silently hang in the loading skeleton forever instead of erroring.

Possible mitigation: a project-level `beforeEach` here defaulting `mocked(fetchPrices).mockResolvedValue({})`, which per-meta `beforeEach` (like the one added in `fiatValue.stories.tsx`) would still override. Small addition, but maybe out of scope for this PR — could also just be tracked as a follow-up issue if you'd rather not expand this change.

Gabi 🤖

gndelia: I don't think this is wrong for now - no other component uses prices, and after all, we won't want any story to call real prices, so having this by default looks fine to me. We can update it in the future if needed.

gabmontes (approved):

PR #797, opened 2026-09-02
Handle unexpected SDK error events in forms

### LINKED ISSUE #793: Swap, Bridge, and Borrow forms don't handle the unexpected-error SDK event
Same root cause as #789/#790, one layer up: every SDK action emits a generic `unexpected-error` from its outer `catch`, and these 11 components track their own `flowStatus` via `onEmitter` but never listen for it — so an unexpected error leaves the form stuck on its pending step forever.

**Affected:**
- `swapForm/deposit.tsx`, `oneStepRedeem.tsx`, `twoStepRedeem.tsx`, `redeemQueue.tsx`, `cancelRedeemModal.tsx`
- `bridgeForm/index.tsx`
- `borrow/borrowForm.tsx`, `borrowMoreForm.tsx`, `repayLoanForm.tsx`, `supplyCollateralForm.tsx`, `withdrawCollateralForm.tsx`

**Possible fix:** one option is adding `emitter.on("unexpected-error", ...)` in each, mapping to that flow's error status — similar to the `markFailed` pattern in `useCancelWithdraw`/`useClaimWithdraw`/`useClaimWithdrawBatch`. Open to other approaches.

Trynax:
### Description

Adds `unexpected-error` event handling to the Swap, Bridge, and Borrow forms. Unexpected SDK failures now transition each flow to its existing error state instead of leaving the UI stuck in a pending state.

### Screenshots

N/A — no visual changes.

### Related issue(s)

Closes #793

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/web/src/components/borrow/borrowForm.tsx
+++ b/web/src/components/borrow/borrowForm.tsx
@@ -215,6 +215,23 @@ export function BorrowForm({
     collateralAmount: collateralAmountBigInt,
     marketId,
     onEmitter(emitter) {
+      let supplyCollateralSucceeded = false;
+
+      const handleFailure = function (
+        status: "borrow-error" | "supply-collateral-error",
+      ) {
+        onFailed();
+        setFlowStatus(status);
+      };
+
+      const handleUnexpectedError = function () {
+        handleFailure(
+          supplyCollateralSucceeded
+            ? "borrow-error"
+            : "supply-collateral-error",
+        );
+      };
+
       emitter.on("user-signed-approval", () => setFlowStatus("approving"));
       emitter.on("approve-transaction-succeeded", () =>
         setFlowStatus("approved"),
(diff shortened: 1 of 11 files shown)


gabmontes (dismissed):
LGTM

web/src/components/swapForm/oneStepRedeem.tsx
```diff
@@ -203,6 +203,10 @@ export function OneStepRedeem({
         onFailed();
         setFlowStatus("redeem-error");
       });
+      emitter.on("unexpected-error", function () {
```
ArturDolzan: While we're here, might be worth a quick look at `redeem-failed` / `request-redeem-failed`. The SDK emits them when `waitForTransactionReceipt` rejects, and these three forms don't listen, so the drawer could still hang.

web/src/components/borrow/borrowForm.tsx
```diff
@@ -274,6 +274,10 @@ export function BorrowForm({
         onFailed();
         setFlowStatus("borrow-error");
       });
+      emitter.on("unexpected-error", function () {
```
ArturDolzan: Wonder if we should double-check the multi-step case here. `unexpected-error` can fire before anything is signed (the allowance read is inside the same try), and `borrow-error` renders the supply step as completed. Might read as "collateral supplied".

web/src/components/swapForm/redeemQueue.tsx
```diff
@@ -181,6 +181,10 @@ function ActiveRedeemDrawer({
         onFailed();
         setFlowStatus("redeem-error");
       });
+      emitter.on("unexpected-error", function () {
```
ArturDolzan: Not sure, but maybe a single `handleFailure` closure per `onEmitter` would read a bit lighter now that the same two lines appear four times in each form?

ArturDolzan (approved):

gabmontes (approved):

gndelia (approved):

PR #798, opened 2026-09-02
Add swap cooldown-enabled command to the CLI

gndelia:
### Description

Adds `vetro-cli swap cooldown-enabled --gateway <addr>`, the read that was still missing from the swap gateway-level reads. It calls `getWithdrawalDelayEnabled` and prints the flag as a JSON boolean.

`swap cooldown` already reads the same flag, but folds it into its result: it prints `0` when the delay is off. An agent that must choose between the one-step and the two-step redeem needs the flag on its own, because a 0-second delay and a disabled queue are different states.

### Screenshots

N/A — no UI changes.

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.


packages/cli/README.md
```diff
@@ -102,6 +102,7 @@ This is the list of commands available
...
 | `vetro-cli swap allowance --token <tok> --account <addr>`            | `allowance`                                                                   | Gateway's spending allowance, in human units. `<tok>` may be a whitelisted token (swap-in) or a pegged token (swap-out; the gateway pulls the pegged token at `send-to-queue`, or at the redeem itself when it is one step) |
 | `vetro-cli swap cooldown --gateway <addr>`                           | `getWithdrawalDelayEnabled`, `getWithdrawalDelay`                             | Redeem Queue cooldown of the gateway, in seconds. `0` when the delay is disabled — a redeem is then one-step, with no queue. This is the gateway-wide contract value, not a per-account one.                                |
+| `vetro-cli swap cooldown-enabled --gateway <addr>`                   | `getWithdrawalDelayEnabled`                                                   | Whether the gateway's Redeem Queue is on. `false` means a redeem is one-step for every account, and `swap send-to-queue` reverts.                                                                                           |
```
ArturDolzan: Perhaps it's worth checking the wording here. `send-to-queue` seems to fail locally before emitting any calldata (the queue-disabled guard in `sendToQueue.ts`), so "fails" might match the write-ops row better than "reverts".

packages/cli/test/e2e/cooldownEnabled.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  beforeAll(async function () {
+    const { publicClient } = createClients(rpcUrl);
+    enabledBefore = await getWithdrawalDelayEnabled(publicClient, {
+      address: gateway,
+    });
+  });
+
+  afterAll(() =>
+    setWithdrawalDelay({ enabled: enabledBefore, gateway, rpcUrl }),
+  );
```
ArturDolzan: Wonder if `setRedeemQueueEnabled` from the helpers would fit here. It already captures the previous state and hands back a restore, the way `sendToQueue.test.ts` and `swap.test.ts` use it. `cooldown.test.ts` needs the manual version only because it also restores the delay.

packages/cli/README.md
gndelia: Fixed in [eb2575df](https://github.com/vetro-protocol/vetro-monorepo/pull/798/commits/eb2575df). The row now says `fails`, to match the write-ops row: the guard throws locally, so nothing reaches the chain.

packages/cli/test/e2e/cooldownEnabled.test.ts
gndelia: Fixed in [6c400b72](https://github.com/vetro-protocol/vetro-monorepo/pull/798/commits/6c400b72). The test uses `setRedeemQueueEnabled` now, the same as `sendToQueue.test.ts` and `swap.test.ts`.

ArturDolzan (approved):

gabmontes (approved):

PR #799, opened 2026-09-03
Allow supply and borrow on existing positions

### LINKED ISSUE #501: Positions on borrow page doesn't allow to supply and borrow

The current list of actions over an open position on the borrow page is listed [here in the manageButton.tsx](web/src/components/borrow/manageButton.tsx). For open positions, you can only 

- add more collateral
- Withdraw collateral
- Borrow more
- Repay loan

Sometimes, in order to borrow more, you need extra collateral, but there's no option to do so. The operation of supplying collateral and borrowing within the same transaction only happens when creating a position from the market page. We should add this action as well for existing positions

The [useSupplyAndBorrow](web/src/hooks/borrow/useSupplyAndBorrow.ts) hook contains how to supply and borrow using the morpho SDK - we can probably reuse something from it (or use it as inspiration) but for existing positions.

The UI should show the [borrowForm](borrowForm.tsx) as part of a drawer, just like other borrow actions do (See [supplyCollateralForm.tsx](web/src/components/borrow/supplyCollateralForm.tsx) for example)

Trynax:
### Description

Adds a “Supply collateral and borrow” action for existing borrow positions.

The flow reuses the borrow form, includes existing position data in the review and borrow limit, and retries only the borrow step if collateral was already supplied. This prevents collateral from being supplied twice.

### Screenshots
image

image

### Related issue(s)

Closes #501

### Checklist

- [x] Manual testing passed.
- [x] Automated tests added.

### DIFF
--- a/web/src/components/borrow/borrowForm.tsx
+++ b/web/src/components/borrow/borrowForm.tsx
@@ -243,7 +243,11 @@ export function BorrowForm({
 
   const handleBorrowFailure = function () {
     onFailed();
-    setFlowStatus("borrow-error");
+    setFlowStatus(
+      supplyCollateralSucceeded.current
+        ? "borrow-error"
+        : "supply-collateral-error",
+    );
   };
 
   const borrowMutation = useSupplyAndBorrow({
--- a/web/src/components/borrow/borrowForm.tsx
+++ b/web/src/components/borrow/borrowForm.tsx
@@ -362,6 +362,12 @@ export function BorrowForm({
   function handleSubmit(e: FormEvent) {
     e.preventDefault();
     if (!inputError) {
+      if (supplyCollateralSucceeded.current) {
+        handleRetry();
+        onDrawerOpenChange(true);
+        return;
+      }
+
       supplyCollateralSucceeded.current = false;
       setStartedWithApproval(!!needsApproval);
       setFlowStatus(needsApproval ? "approving" : "supply-collateral-ready");
--- a/web/src/components/borrow/borrowForm.tsx
+++ b/web/src/components/borrow/borrowForm.tsx
@@ -129,6 +129,12 @@ function getInputError({
   return undefined;
 }
 
+const getActivityType = (position?: AccrualPosition) =>
+  position ? "borrow-more" : "open-position";
+
+const getInputContainerClassName = (position?: AccrualPosition) =>
+  `flex flex-col gap-1 ${position ? "border-t border-gray-200 p-6" : "p-2 md:px-1.5 xl:px-2"}`;
+
 type Props = {
   borrowInput: string;
   collateralInput: string;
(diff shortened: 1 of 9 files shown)


web/src/components/borrow/borrowForm.tsx
```diff
@@ -141,6 +149,8 @@ export function BorrowForm({
   onBorrowChange,
   onCollateralChange,
   onDrawerOpenChange,
+  onSuccess,
+  position,
```
ArturDolzan, line 159: Perhaps it's worth checking the activity entry for this new flow. Since the form is now reused on an existing position, the tracked title still resolves to "Open VUSD loan". Wonder if `activity.borrow-more-title` would read better there.

web/src/components/borrow/supplyCollateralAndBorrowDrawerForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <Drawer onClose={onClose}>
+      <div className="flex h-full flex-col">
+        <DrawerTitle>
+          {t("pages.borrow.supply-collateral-and-borrow")}
+        </DrawerTitle>
+        {position ? (
+          <div className="flex-1 overflow-y-auto">
+            <Suspense fallback={<DrawerLoader />}>
+              <BorrowForm
```
ArturDolzan: Might be worth a quick look at the horizontal padding here. The other Manage drawers use `p-6` / `DrawerFeesContainer` (`px-6`) to match `DrawerTitle`, while the reused form keeps its sidebar `p-2 md:px-1.5`. The Updated column looks a bit tight against the edge in the screenshot.

web/src/components/borrow/borrowForm.tsx
```diff
@@ -277,6 +295,40 @@ export function BorrowForm({
     },
   });
 
+  const handleBorrowMoreSuccess = function () {
```
ArturDolzan: Wonder if we could reuse the new handlers for `borrowMutation` too. They look identical to the five inline blocks just above, so hoisting them (and dropping the `More` from the name) would keep both paths in sync.

web/src/components/borrow/borrowForm.tsx
```diff
@@ -231,9 +248,10 @@ export function BorrowForm({
       emitter.on("user-signed-supply-collateral", () =>
         setFlowStatus("supplying-collateral"),
       );
-      emitter.on("supply-collateral-transaction-succeeded", () =>
-        setFlowStatus("supplied-collateral"),
-      );
+      emitter.on("supply-collateral-transaction-succeeded", function () {
+        supplyCollateralSucceeded.current = true;
+        setFlowStatus("supplied-collateral");
+      });
```
gabmontes, line 277: On `borrowMutation`'s `onEmitter`, there's no handler for `"unexpected-error"`, unlike the new `borrowMoreMutation` below which does handle it. `supplyCollateralAndBorrow` emits this on any uncaught exception (e.g. `getMarketParams`/`allowance` calls in the core supply/borrow steps aren't wrapped in `.catch()`, so an RPC hiccup bubbles up here), so if it fires the drawer stays stuck on `"borrowing"`/`"supplying-collateral"` forever with no way to recover. Pre-existing gap, but worth closing while this block is being touched.

— Gabi 🤖

web/src/components/borrow/borrowForm.tsx
```diff
@@ -252,28 +289,31 @@ export function BorrowForm({
...
-      });
-      emitter.on("user-signing-borrow-assets-error", function () {
-        onFailed();
-        setFlowStatus("borrow-error");
+      emitter.on("borrow-assets-transaction-succeeded", handleBorrowSuccess);
+      emitter.on("borrow-assets-transaction-reverted", handleBorrowFailure);
+      emitter.on("borrow-assets-failed", handleBorrowFailure);
+      emitter.on("borrow-assets-failed-validation", handleBorrowFailure);
+      emitter.on("user-signing-borrow-assets-error", handleBorrowFailure);
+      emitter.on("unexpected-error", handleBorrowFailure);
```
ArturDolzan, line 301: Following up on Gabi's `unexpected-error` thread. Might be worth checking where that error can come from: `runSupplyCollateralCore` awaits `getMarketParams`/`allowance` uncaught, so an RPC blip in the supply leg lands on `borrow-error`, and `getSupplyCollateralStepStatus` renders that step as completed. Could we branch on `supplyCollateralSucceeded.current`?

web/src/components/borrow/borrowForm.tsx
```diff
@@ -199,17 +222,30 @@ export function BorrowForm({
     collateralToken,
     marketId,
   });
+  const activityType = getActivityType(position);
 
   const { onCompleted, onFailed, onPending, onTransactionHash } =
     useActivityTracking({
       page: "borrow",
-      text: t("pages.borrow.activity.open-position-text", {
+      text: t(`pages.borrow.activity.${activityType}-text`, {
```
ArturDolzan, line 230: The locale orphaned-key check seems to flag `pages.borrow.activity.open-position-text` and `-title` now that they are only built via the template key.

Image

ArturDolzan (changes requested):

gndelia (changes requested):
@Trynax make sure your commits are signed

image


They should all appear as "Verified"

web/src/components/borrow/borrowForm.tsx
```diff
@@ -291,16 +335,34 @@ export function BorrowForm({
...
+      borrowMoreMutation.mutate();
+      return;
+    }
     borrowMutation.mutate();
   };
 
   function handleSubmit(e: FormEvent) {
     e.preventDefault();
     if (!inputError) {
+      supplyCollateralSucceeded.current = false;
```
ArturDolzan, line 371: Perhaps it's worth checking the path where the supply succeeds, the borrow fails, and the user dismisses the progress drawer instead of clicking Retry. The form stays mounted here, so submitting again seems to run the supply leg a second time.

PR #805, opened 2026-09-04
Reconcile pending activity on reload

### LINKED ISSUE #804: Reconcile pending transaction activity after page reload
### Description

Pending transaction activities are saved in local storage, but their status is only updated through in-memory transaction callbacks.

If the page is reloaded while a transaction is pending, the transaction may later succeed or revert on-chain, while the activity remains pending in the wallet drawer indefinitely.

### Steps to reproduce

1. Start the local fork:

   ```bash
   pnpm --filter @vetro-protocol/web dev:fork
   ```

2. Disable automining:

   ```bash
   cast rpc evm_setAutomine false --rpc-url http://127.0.0.1:8545
   ```

3. Submit a transaction that does not require approval, or use an already-approved token.
4. Confirm the activity appears as pending.
5. Hard reload the page.
6. Mine the transaction:

   ```bash
   cast rpc evm_mine --rpc-url http://127.0.0.1:8545
   ```

7. Open the wallet activity again.

The transaction succeeds on-chain, but the activity still shows as pending.

### Environment

- **OS**: Linux
- **Runtime/Browser/Version**: Node 24, Chrome
- **Wallet**: Vetro local dev wallet (`dev:fork`)

### Proposed fix

On application startup and while the application is open, check persisted pending non-bridge activities against their transaction receipts.

Successful transactions should become completed, reverted transactions should become failed, and unmined transactions or temporary RPC failures should remain pending.

Trynax:
### Description

Pending transaction activities are saved in local storage, but their status was only updated by callbacks kept in memory. After a page reload, a mined or reverted non-bridge transaction could remain pending indefinitely.

### Screenshots

N/A — no visual layout changes.

### Related issue(s)

Closes #804 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/app.tsx
+++ b/web/src/app.tsx
@@ -29,6 +29,8 @@ import {
 import { featureFlags } from "utils/featureFlags";
 import { isGeoRestricted } from "utils/geoRestriction";
 
+import { usePendingActivityReconciliation } from "./hooks/usePendingActivityReconciliation";
+
 const SentryRoutes = Sentry.withSentryReactRouterV7Routing(Routes);
 
 const analyticsEnabled =
@@ -141,22 +143,26 @@ function LanguageRoutes() {
   );
 }
 
-export const App = () => (
-  <BrowserRouter>
-    <NuqsAdapter>
-      {analyticsEnabled && <AnalyticsTracker />}
-      <AppViewport>
-        <SentryRoutes>
-          {/* Redirect root to English */}
-          <Route element={<Navigate replace to="/en" />} path="/" />
+export function App() {
+  usePendingActivityReconciliation();
 
-          {/* Language-prefixed routes */}
-          <Route element={<LanguageRoutes />} path="/:lang/*" />
+  return (
+    <BrowserRouter>
+      <NuqsAdapter>
+        {analyticsEnabled && <AnalyticsTracker />}
+        <AppViewport>
+          <SentryRoutes>
+            {/* Redirect root to English */}
+            <Route element={<Navigate replace to="/en" />} path="/" />
 
-          {/* Catch-all: redirect unknown paths to English */}
-          <Route element={<Navigate replace to="/en" />} path="*" />
-        </SentryRoutes>
-      </AppViewport>
-    </NuqsAdapter>
-  </BrowserRouter>
-);
+            {/* Language-prefixed routes */}
+            <Route element={<LanguageRoutes />} path="/:lang/*" />
+
+            {/* Catch-all: redirect unknown paths to English */}
+            <Route element={<Navigate replace to="/en" />} path="*" />
+          </SentryRoutes>
+        </AppViewport>
+      </NuqsAdapter>
+    </BrowserRouter>
+  );
+}
--- a/web/src/app.tsx
+++ b/web/src/app.tsx
@@ -29,8 +29,6 @@ import {
 import { featureFlags } from "utils/featureFlags";
 import { isGeoRestricted } from "utils/geoRestriction";
 
-import { usePendingActivityReconciliation } from "./hooks/usePendingActivityReconciliation";
-
 const SentryRoutes = Sentry.withSentryReactRouterV7Routing(Routes);
 
 const analyticsEnabled =
@@ -143,26 +141,22 @@ function LanguageRoutes() {
   );
 }
 
-export function App() {
-  usePendingActivityReconciliation();
+export const App = () => (
+  <BrowserRouter>
+    <NuqsAdapter>
+      {analyticsEnabled && <AnalyticsTracker />}
+      <AppViewport>
+        <SentryRoutes>
+          {/* Redirect root to English */}
+          <Route element={<Navigate replace to="/en" />} path="/" />
 
-  return (
-    <BrowserRouter>
-      <NuqsAdapter>
-        {analyticsEnabled && <AnalyticsTracker />}
-        <AppViewport>
-          <SentryRoutes>
-            {/* Redirect root to English */}
-            <Route element={<Navigate replace to="/en" />} path="/" />
+          {/* Language-prefixed routes */}
+          <Route element={<LanguageRoutes />} path="/:lang/*" />
 
-            {/* Language-prefixed routes */}
-            <Route element={<LanguageRoutes />} path="/:lang/*" />
-
-            {/* Catch-all: redirect unknown paths to English */}
-            <Route element={<Navigate replace to="/en" />} path="*" />
-          </SentryRoutes>
-        </AppViewport>
-      </NuqsAdapter>
-    </BrowserRouter>
-  );
-}
+          {/* Catch-all: redirect unknown paths to English */}
+          <Route element={<Navigate replace to="/en" />} path="*" />
+        </SentryRoutes>
+      </AppViewport>
+    </NuqsAdapter>
+  </BrowserRouter>
+);
--- a/web/src/hooks/usePendingActivityReconciliation.ts
+++ b/web/src/hooks/usePendingActivityReconciliation.ts
@@ -1,12 +1,15 @@
-import { useEffect, useMemo } from "react";
+import { useEffect, useMemo, useRef } from "react";
+import { getTransactionReceipt } from "viem/actions";
 import { useAccount } from "wagmi";
 
 import { updateActivity, useActivities } from "../stores/activityStore";
+import { SECONDS_PER_DAY, unixNowTimestamp } from "../utils/date";
 import { getPendingActivityStatus } from "../utils/reconcilePendingActivity";
 
 import { useEthereumClient } from "./useEthereumClient";
 
 const pendingActivityPollInterval = 10_000;
+const pendingActivityMaxAge = SECONDS_PER_DAY;
 
 export function usePendingActivityReconciliation() {
   const { address } = useAccount();
@@ -17,12 +20,22 @@ export function usePendingActivityReconciliation() {
   // does not mean that the bridged funds have arrived. They need separate
   // delivery tracking.
   const pendingActivities = useMemo(
-    () =>
-      activities.filter(
+    function getPendingActivities() {
+      const now = unixNowTimestamp();
+      return activities.filter(
         (activity) =>
-          activity.status === "pending" && activity.page !== "bridge",
-      ),
+          activity.status === "pending" &&
+          activity.page !== "bridge" &&
+          now - activity.date < pendingActivityMaxAge,
+      );
+    },
     [activities],
+  );
+  const pendingActivitiesRef = useRef(pendingActivities);
+  pendingActivitiesRef.current = pendingActivities;
+  const pendingActivityHashes = useMemo(
+    () => pendingActivities.map(({ txHash }) => txHash).join(","),
+    [pendingActivities],
   );
 
   useEffect(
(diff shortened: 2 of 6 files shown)


web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,77 @@
...
+        () => void reconcile(),
+        pendingActivityPollInterval,
+      );
+
+      return function cleanup() {
+        isActive = false;
+        clearInterval(interval);
+      };
+    },
+    [address, client, pendingActivities],
```
ArturDolzan: Might be worth a quick look at the `pendingActivities` dep here. The store returns a fresh array on every write, so each `updateActivity` seems to tear the effect down, drop the receipts already fetched in that run (`isActive = false`) and restart the interval. Wonder if a key built from the hashes would be steadier.

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,77 @@
...
+  const client = useEthereumClient();
+
+  // Bridge transactions can start on several chains, and source confirmation
+  // does not mean that the bridged funds have arrived. They need separate
+  // delivery tracking.
+  const pendingActivities = useMemo(
+    () =>
+      activities.filter(
+        (activity) =>
+          activity.status === "pending" && activity.page !== "bridge",
```
ArturDolzan: Perhaps it's worth checking what happens with a transaction the user sped up or cancelled in their wallet. The original hash never gets a receipt, so it seems we'd keep polling it every 10s forever across every session. Wonder if bounding it by `activity.date` would help.

Trynax: Good point @ArturDolzan  should we handle cancelled or replaced transactions here, or open a separate issue for them?

ArturDolzan: I'd say a separate issue for the replaced/cancelled part. telling "replaced" from "still pending" needs the nonce stored on the activity, and viem's onReplaced won't help us after a reload since the original tx is already gone from the mempool. But the endless polling I'd fix here... any hash that never gets a receipt sticks around forever, and we never prune activities, so every future session re-polls it. @gndelia any thoughts on this?

Trynax: Makes sense. I’ll keep replaced/cancelled transactions separate and add a date-based cutoff for stale pending activities. Is 24 hours a reasonable timeout?

ArturDolzan: 24h works for me.. nodes usually drop txs from the mempool well before that, so anything still unresolved after a day isn't coming back. Two small things while you're in there: worth putting the cutoff in the `pendingActivities` filter rather than inside the util, so the interval doesn't even start once everything's stale. And mind the unit.. `activity.date` is seconds, but `pendingActivityPollInterval` right above it is ms.

web/src/hooks/usePendingActivityReconciliation.ts
gabmontes: `pendingActivities` gets a new array reference on every store write for this address, not just when the pending set actually changes — `write()` in `activityStore.ts` always caches a fresh array, and `.filter()` here always returns a new one too. So e.g. an unrelated bridge activity update (still excluded by the filter) tears down and restarts this effect (clear interval, immediate `reconcile()`, new interval) even though nothing relevant changed. Consider keying the effect off a stable derived value (e.g. the set of pending tx hashes) instead of the array reference.

Gabi 🤖

Trynax: I had already addressed this in the latest update before seeing your comment 😅

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,101 @@
...
+  // does not mean that the bridged funds have arrived. They need separate
+  // delivery tracking.
+  const pendingActivities = useMemo(
+    function getPendingActivities() {
+      const now = unixNowTimestamp();
+      return activities.filter(
+        (activity) =>
+          activity.status === "pending" &&
+          activity.page !== "bridge" &&
+          now - activity.date < pendingActivityMaxAge,
```
ArturDolzan: Perhaps it's worth checking someone who reopens the app more than 24h after submitting. The cutoff is on when the tx was sent rather than on how long we've been polling it, so the activity gets filtered out before its first check — a tx that did get mined stays pending forever, and pending items always pass the drawer filter (`walletDrawerContent.tsx:70`), so there's no way to clear it either. Wonder if letting each hash through once per session before applying the cutoff would cover it.

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,101 @@
...
+  const { address } = useAccount();
+  const activities = useActivities(address);
+  const client = useEthereumClient();
+
+  // Bridge transactions can start on several chains, and source confirmation
+  // does not mean that the bridged funds have arrived. They need separate
+  // delivery tracking.
+  const pendingActivities = useMemo(
+    function getPendingActivities() {
+      const now = unixNowTimestamp();
```
ArturDolzan: Might be worth a quick look at when the cutoff gets re-evaluated. `now` is captured in the memo, which only recomputes when `activities` changes identity — and a hash that never resolves produces no store writes by definition, so on a tab left open for days it seems we'd keep polling past the 24h. Wonder if recomputing `now` inside `reconcile` would close it.

Trynax: @ArturDolzan  addressed both comments, stale pending hashes are checked once after reload, and the cutoff is re-evaluated on each reconciliation cycle

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,140 @@
...
+              },
+            ),
+          );
+
+          if (isActive) {
+            reconciledActivities.forEach(function updateReconciledActivity({
+              activity,
+              status,
+            }) {
+              checkedHashesRef.current.add(activity.txHash);
```
ArturDolzan, line 100: Perhaps it's worth checking whether a failed RPC call should count as a "check" here. It looks like `catch` returns `undefined` for both "receipt not found" and a network error, and we add the hash to `checkedHashes` either way, so one blip could spend a stale activity's only lookup for the session.

web/src/app.tsx
```diff
@@ -141,22 +143,26 @@ function LanguageRoutes() {
...
-export const App = () => (
-  <BrowserRouter>
-    <NuqsAdapter>
-      {analyticsEnabled && <AnalyticsTracker />}
-      <AppViewport>
-        <SentryRoutes>
-          {/* Redirect root to English */}
-          <Route element={<Navigate replace to="/en" />} path="/" />
+export function App() {
+  usePendingActivityReconciliation();
```
ArturDolzan, line 147: Might be worth a quick look at mounting this as a small null-rendering component (like `AnalyticsTracker` / `AddressRestriction`) instead of hooking it into `App`. As-is, every activity store write re-renders the whole route tree, which `App` never did before.

web/test/utils/reconcilePendingActivity.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    expect(
+      getPendingActivitiesToReconcile({
+        activities: [activity],
+        checkedHashes: new Set(),
+        maxAge,
+        now,
+      }),
+    ).toEqual([activity]);
+  });
```
ArturDolzan, line 38: Wonder if we should double-check this one. With `date: now - maxAge` the age check is `100 < 100` (false), so it only passes via the empty `checkedHashes`. Maybe a checked hash plus `now - maxAge + 1` would actually exercise the freshness branch?

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,142 @@
...
+        return undefined;
+      }
+
+      const account = address;
+      const publicClient = client;
+      let isActive = true;
+      let isChecking = false;
+      const intervalRef = {
+        current: undefined as ReturnType<typeof setInterval> | undefined,
+      };
```
ArturDolzan: Perhaps it's worth checking whether a plain `let intervalId` would read a bit cleaner than the `{ current }` wrapper here. It sits right next to `isActive`/`isChecking`, and the closure captures it the same way.

web/src/utils/reconcilePendingActivity.ts
```diff
@@ -0,0 +1,58 @@
...
+  maxAge: number;
+  now: number;
+}) =>
+  // Give persisted activities one receipt lookup after a new session starts,
+  // even when they are already older than the polling cutoff.
+  activities.filter(
+    (activity) =>
+      activity.status === "pending" &&
+      activity.page !== "bridge" &&
+      (now - activity.date < maxAge || !checkedHashes.has(activity.txHash)),
```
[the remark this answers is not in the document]
Trynax: @ArturDolzan, the 24-hour cutoff is intentional to avoid polling stale or replaced transactions forever, based on the earlier discussion.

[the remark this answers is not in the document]
ArturDolzan: Agreed on the cutoff, settled from my side. Small hole though: only `TransactionReceiptNotFoundError` returns `null`, so any other RPC blip gives `undefined`, the hash never reaches `checkedHashes`, and the filter keeps it regardless of age... the exact thing the 24h is meant to stop. Should a failed lookup count as spent past `maxAge`?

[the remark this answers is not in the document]
Trynax: @ArturDolzan Yes, that makes sense. fixed it so fresh RPC errors remain retryable, while stale activities are marked as checked after 24 hours.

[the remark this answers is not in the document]
ArturDolzan: Looks good. Thanks!

ArturDolzan (dismissed):

web/src/utils/reconcilePendingActivity.ts
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  activity: Activity;
+  getReceipt: ReceiptReader;
+}): Promise<"completed" | "failed" | null | undefined> {
+  if (activity.status !== "pending" || activity.page === "bridge") {
+    return undefined;
+  }
+
+  try {
+    const receipt = await getReceipt(activity.txHash as Hash);
```
gndelia: I don't think we need `getReceipt` as a parameter here. IT will always be `getTransactionReceipt(publicClient, { hash }),` from `viem/actions`. I assume you did so for testing; If that's the case, let's drop it (so the API is cleaner - you can just accept one parameter - the activity) and use mocking in the tests

gndelia:
_This approach is not totally incorrect_, but I feel the long-term problem to be solved is that transactions should not be stored in memory. However, for the time being, this is ok. I did leave you some comments to address, though. Thanks for the contribution!

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,145 @@
...
+  );
+
+  useEffect(
+    function reconcilePendingActivities() {
+      if (!address || !client || pendingActivities.length === 0) {
+        return undefined;
+      }
+
+      const account = address;
+      const publicClient = client;
```
gndelia: better name `publicClient` directly on line 21 ?

web/src/hooks/usePendingActivityReconciliation.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      return function cleanup() {
+        isActive = false;
+        if (intervalId) {
+          clearInterval(intervalId);
+        }
+      };
+    },
+    [address, client, pendingActivityHashes],
+  );
```
gndelia: A trick you can use: combine `useQueries` from `react-query` and write a `queryFn` that calls `updateActivity` for each activity. 

This way, you can take advantage of the whole polling and drop functionality that `react-query` offers you. And no `useEffect` either!

Trynax:
> _This approach is not totally incorrect_, but I feel the long-term problem to be solved is that transactions should not be stored in memory. However, for the time being, this is ok. I did leave you some comments to address, though. Thanks for the contribution!

Thanks @gndelia, I’ve addressed your comments locally and will push the changes soon. Should I keep this as a focused fix and leave the longer-term transaction-state approach for a separate PR, or would you prefer me to rework this PR around that approach?

gndelia:
> > _This approach is not totally incorrect_, but I feel the long-term problem to be solved is that transactions should not be stored in memory. However, for the time being, this is ok. I did leave you some comments to address, though. Thanks for the contribution!
> 
> Thanks @gndelia, I’ve addressed your comments locally and will push the changes soon. Should I keep this as a focused fix and leave the longer-term transaction-state approach for a separate PR, or would you prefer me to rework this PR around that approach?

No, let's keep this approach. I need to think how to implement the longer-term one. Thanks !

PR #812, opened 2026-09-09
Add max repayment to borrow flow

### LINKED ISSUE #800: Add a max button to the repay loan form

Add a MAX button to the repay loan form to fill the maximum valid repayment amount.

The form currently requires users to enter the repayment amount manually. There is no quick way to repay the full outstanding debt, and manually entered values may become outdated as interest accrues.

### Proposed implementation

Use the current outstanding debt and loan-token balance to calculate the maximum valid amount, then set the input to the lower of those two values.

Trynax:
### Description

Adds a MAX button to the repay loan form. MAX fills the lower of the current debt and the wallet’s loan-token balance.

Near-full repayments use Morpho’s share-based repayment path to avoid leaving debt dust as interest accrues. The successful repayment cache update also uses mutually exclusive asset/share arguments.

### Screenshots

image

### Related issue(s)

Closes #800

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/components/borrow/repayLoanForm.tsx
+++ b/web/src/components/borrow/repayLoanForm.tsx
@@ -5,6 +5,7 @@ import { getChainAddresses } from "@morpho-org/blue-sdk";
 import { Button } from "components/base/button";
 import { RenderCryptoValue } from "components/base/cryptoValue";
 import { RenderFiatValue } from "components/base/fiatValue";
+import { MaxButton } from "components/base/maxButton";
 import { Toast } from "components/base/toast";
 import {
   type Step,
@@ -37,7 +38,14 @@ import { useTranslation } from "react-i18next";
 import { formatLtvAsPercentage } from "utils/borrowReview";
 import { formatNumber } from "utils/format";
 import { isGeoRestricted } from "utils/geoRestriction";
+import {
+  getCurrentBorrowAssets,
+  getMaxRepayable,
+  getRepayApprovalAmount,
+  getRepayShares,
+} from "utils/repay";
 import { parseTokenUnits } from "utils/token";
+import { formatUnits } from "viem";
 import { useAccount } from "wagmi";
 
 import { PositionReview } from "./positionReview";
@@ -252,25 +260,45 @@ export function RepayLoanForm({ market, onClose }: Props) {
 
   const { data: nativeBalanceData } = useNativeBalance(ethereumChain.id);
 
-  const { data: needsApproval } = useNeedsApproval({
-    amount: repayAmountBigInt,
-    spender: getChainAddresses(ethereumChain.id).morpho,
-    token: loanToken,
-  });
-
   const { data: positionInfo } = usePositionInfo(marketId);
 
   const { data: morphoMarket } = useMorphoMarket(marketId);
 
-  const currentBorrowAssets =
-    positionInfo && morphoMarket
-      ? morphoMarket.toBorrowAssets(positionInfo.borrowShares)
-      : undefined;
+  const currentBorrowAssets = getCurrentBorrowAssets({
+    borrowShares: positionInfo?.borrowShares,
+    market: morphoMarket,
+  });
+
+  const maxRepayable = getMaxRepayable({
+    borrowShares: positionInfo?.borrowShares,
+    loanTokenBalance: loanBalance,
+    market: morphoMarket,
+  });
+
+  const repayShares = getRepayShares({
+    amount: repayAmountBigInt,
+    borrowAssets: currentBorrowAssets,
+    borrowShares: positionInfo?.borrowShares,
+    loanTokenBalance: loanBalance,
+  });
+
+  const repayApprovalAmount = getRepayApprovalAmount({
+    amount: repayAmountBigInt,
+    loanTokenBalance: loanBalance,
+    shares: repayShares,
+  });
+
+  const { data: needsApproval } = useNeedsApproval({
+    amount: repayApprovalAmount,
+    spender: getChainAddresses(ethereumChain.id).morpho,
+    token: loanToken,
+  });
 
   const networkFee = useTotalRepayFees({
     amount: repayAmountBigInt,
-    approveAmount: undefined,
+    approveAmount: repayShares !== undefined ? repayApprovalAmount : undefined,
     marketId,
+    shares: repayShares,
     token: loanToken,
   });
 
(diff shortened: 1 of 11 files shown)


web/src/utils/repay.ts
```diff
@@ -0,0 +1,85 @@
...
+}) {
+  const isNearFullRepayment =
+    amount > 0n &&
+    borrowAssets !== undefined &&
+    borrowAssets > 0n &&
+    borrowShares !== undefined &&
+    borrowShares > 0n &&
+    loanTokenBalance !== undefined &&
+    loanTokenBalance >= borrowAssets &&
+    amount * BPS_DENOMINATOR >= borrowAssets * FULL_REPAYMENT_THRESHOLD_BPS;
```
ArturDolzan, line 41: Perhaps it's worth double-checking the 99% threshold. If someone types 99.5% of their debt on purpose, we'd repay 100% and pull more than "You will repay" shows. Wonder if triggering shares only at `amount >= borrowAssets` would be safer?

web/src/utils/repay.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const getRepayApprovalAmount = ({
+  amount,
+  loanTokenBalance,
+  shares,
+}: {
+  amount: bigint;
+  loanTokenBalance: bigint | undefined;
+  shares: bigint | undefined;
+}) => (shares === undefined ? amount : (loanTokenBalance ?? amount));
```
ArturDolzan, line 74: Perhaps it's worth confirming we want to approve the full wallet balance on a share-based repay. A small buffer over the debt, capped at the balance, would cover the accrual drift without leaving a wallet-sized allowance sitting on Morpho.

web/src/utils/repay.ts
Trynax: Nice catch @ArturDolzan I checked the flow and 99% share mode can close the full position while the UI still shows the entered amount as a partial repayment. I’ll restrict shares to an exact full repayment and keep lower amounts asset-based.

web/src/utils/repay.ts
Trynax: The full balance covers interest accrual but can leave a large allowance. I agree we should use the debt plus a small buffer, capped by the balance. Any preferred buffer?

Trynax:
Gm @ArturDolzan,  I went with a 1% (100 bps) buffer over the current debt, capped by the wallet balance. This covers short interest accrual without approving the full wallet balance.

web/src/utils/repay.ts
ArturDolzan: Perhaps it's worth a follow-up on the 1% buffer capped at the balance — that part looks right to me. The thing is `getRepayShares` still arms shares at `loanTokenBalance >= borrowAssets`, so when the wallet holds almost exactly the debt the cap swallows the whole buffer and we're back to zero headroom.

Might be worth a quick look at the case where the wallet balance equals the debt almost exactly. Since Morpho rounds `toBorrowAssets(shares, "Up")` at execution time and we cap the approval at the balance, I think the transfer reverts there. Could we require a few bps of headroom before arming shares?

Trynax: > Perhaps it's worth a follow-up on the 1% buffer capped at the balance — that part looks right to me. The thing is `getRepayShares` still arms shares at `loanTokenBalance >= borrowAssets`, so when the wallet holds almost exactly the debt the cap swallows the whole buffer and we're back to zero headroom.
> 
> Might be worth a quick look at the case where the wallet balance equals the debt almost exactly. Since Morpho rounds `toBorrowAssets(shares, "Up")` at execution time and we cap the approval at the balance, I think the transfer reverts there. Could we require a few bps of headroom before arming shares?

Yeah, that’s true @ArturDolzan. Fixed in the latest commit, share repayment now requires the wallet balance to cover the debt plus the existing 1% buffer, otherwise, it falls back to asset-based repayment.

web/src/utils/repay.ts
```diff
@@ -0,0 +1,104 @@
...
+  const isFullRepayment =
+    amount > 0n &&
+    borrowAssets !== undefined &&
+    borrowAssets > 0n &&
+    borrowShares !== undefined &&
+    borrowShares > 0n &&
+    loanTokenBalance !== undefined &&
+    loanTokenBalance >=
+      borrowAssets + applyBps(borrowAssets, REPAYMENT_BUFFER_BPS) &&
+    amount >= borrowAssets;
```
ArturDolzan: This comparison has zero slack: MAX writes exactly `toBorrowAssets(borrowShares)` and `currentBorrowAssets` is that same value. But the market query re-accrues to now on every refetch (staleTime 0 + refetch on focus), so one wallet popup makes the debt a hair bigger than the frozen input and we quietly fall back to the asset path — the dust we're trying to kill. Could we carry a "user picked MAX" flag instead of comparing against a moving target?

web/src/utils/repay.ts
```diff
@@ -0,0 +1,104 @@
...
+  loanTokenBalance: bigint | undefined;
+}) {
+  const isFullRepayment =
+    amount > 0n &&
+    borrowAssets !== undefined &&
+    borrowAssets > 0n &&
+    borrowShares !== undefined &&
+    borrowShares > 0n &&
+    loanTokenBalance !== undefined &&
+    loanTokenBalance >=
```
ArturDolzan: Coming from the last round: requiring 1.01x the debt means anyone holding between 1.00x and 1.01x can't reach the shares path at all — and that's most "pay off my loan" users. MAX still shows them the full debt with no error, so they repay by assets and keep a residual. What if the buffer lived only in the approval amount and this check stayed at `>= borrowAssets`?

ArturDolzan (commented):
Two notes on the full-repayment gate — both about it turning itself off in cases where we'd want it on 🙂

Trynax:
> Two notes on the full-repayment gate — both about it turning itself off in cases where we'd want it on 🙂

@ArturDolzan Good catches on both. fixed both now, MAX now stays on the share path when interest accrues, and the 1% buffer is used only for approval.

web/src/components/borrow/repayLoanForm.tsx
```diff
@@ -252,25 +261,60 @@ export function RepayLoanForm({ market, onClose }: Props) {
...
+    setIsMaxRepayment(false);
+    onRepayChange(value);
+  };
+
+  const handleMaxRepay = function () {
+    if (maxRepayable === undefined) {
+      return;
+    }
+
+    setIsMaxRepayment(true);
```
ArturDolzan: Perhaps it's worth checking the case where MAX is limited by the wallet balance instead of the debt. We flag it as `isMaxRepayment` anyway, so if the balance later crosses the debt (focus refetch), the share path arms with an approval sized off the old partial amount. Could we capture `maxRepayable >= currentBorrowAssets` at click time?

Trynax: Agreed @ArturDolzan  and fixed MAX now captures whether it covered the full debt at click time, so a wallet-limited MAX stays asset-based after a refetch.

PR #816, opened 2026-09-10
Add retry to stake flows

### LINKED ISSUE #768: Add retry functionality

The stake, both for deposit and withdrawal, has no retry functionality to handle error

Trynax:
### Description

Adds retry handling to Earn staking deposit and withdrawal flows.

Failed approval, deposit, instant withdrawal, and queued withdrawal steps now show a retry action. Retrying resumes from the failed step without repeating successful steps.

### Screenshots

image

### Related issue(s)

Closes #768

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/web/src/hooks/useStakeDeposit.ts
+++ b/web/src/hooks/useStakeDeposit.ts
@@ -5,6 +5,7 @@ import { tokenBalanceQueryKey } from "@hemilabs/react-hooks/useTokenBalance";
 import { useUpdateNativeBalanceAfterReceipt } from "@hemilabs/react-hooks/useUpdateNativeBalanceAfterReceipt";
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { deposit } from "@vetro-protocol/earn/actions";
+import { useRef } from "react";
 import type { TokenWithGateway } from "types";
 import { type CostBases, bumpCostBasis } from "utils/costBasis";
 import type { Address, TransactionReceipt } from "viem";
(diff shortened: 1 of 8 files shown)


packages/earn/src/actions/wallet/deposit.ts
```diff
@@ -198,7 +198,7 @@ const runDeposit = (walletClient: WalletClient, params: DepositParams) =>
       const depositReceipt = await waitForTransactionReceipt(walletClient, {
         hash: depositHash,
       }).catch(function (error: Error) {
-        emitter.emit("deposit-failed", error);
+        emitter.emit("deposit-transaction-unknown", depositHash, error);
```
gndelia: 1. We should keep the events name -were they updated for any reason?
2. The hash is sent but not used, so let's keep the signature and not emit them. It seems we can do the same as others.

Although for these functions retry wasn't implemented, I know it was on others, so check how it was done in other places to replicate the structure

gndelia (changes requested):

packages/earn/src/actions/wallet/deposit.ts
Trynax: yeah that's right @gndelia, I changed the names to track unknown receipts, but the hooks already handle that. I've restored the existing events.

Trynax: Also would like to mention that I sent you a message on X earlier, kindly check it whenever you’re free.

gndelia: @Trynax, I don't use (nor have I enabled) the X chat, so I cannot even see it. If this is asking about open positions at the company, there are none at the moment.

Trynax: @gndelia No worries, it wasn’t about position, I just wanted to connect with you. No problem if you don’t use X.

PR #821, opened 2026-09-11
Add swap preview-mint command to the CLI

gndelia:
### Description

Adds `vetro-cli swap preview-mint --from <tok> --amount <n>`. It infers the gateway from the whitelisted token, reads `previewDeposit`, and prints the pegged token that would be minted as a decimal string.

### Screenshots

N/A — CLI change.

### Related issue(s)

Related to #445

### Checklist

- [x] Manual testing passed.


packages/cli/src/features/swap/commands/previewMint.ts
```diff
@@ -0,0 +1,39 @@
...
+      const { client } = await createVetroClient(
+        command.optsWithGlobals<GlobalOptions>(),
+      );
+      const tokenIn = await resolveWhitelistedToken({
+        client,
+        value: options.from,
+      });
+      const peggedTokenOut = await previewDeposit(client, {
+        address: tokenIn.gatewayAddress,
+        amountIn: parseUnits(options.amount, tokenIn.decimals),
```
ArturDolzan: Perhaps it's worth checking what `--amount 0.0000001` does for a 6 decimal token here. It looks like `parseUnits` rounds it to `0n` and `previewDeposit` then throws the generic "Amount must be greater than 0", while `send-to-queue` gives the nicer "rounds down to 0" message. Might be nice to mirror that?

gndelia: Good catch. Fixed in [32bd05b7](https://github.com/vetro-protocol/vetro-monorepo/pull/821/commits/32bd05b7e09ba9a5a462c816ba01a6b7d0afcdc1): `preview-mint` now guards the parsed amount and prints the same `it rounds down to 0` message as `send-to-queue`, with an e2e case covering it.

Note that `swap mint` and `swap approve` have the same gap. They are outside this PR, so I left them alone — happy to follow up separately.

ArturDolzan (dismissed):

packages/cli/test/e2e/previewMint.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    ...extra,
+    "--rpc-url",
+    rpcUrl,
+  ];
+
+  it("reads the preview by symbol", async function () {
+    const preview = await runCli(
+      previewMintOnFork(["--from", usdc.symbol, "--amount", swapAmount]),
+    );
+    expect(preview).toMatch(/^\d+$/);
```
gabmontes: `toMatch(/^\d+$/)` passes on `"0"` and on any magnitude — including `100e6`, which is exactly what a decimals mixup would produce if the pegged side were read with the whitelisted token's decimals. That's the most likely bug in this command and the assertion can't see it.

`mint.test.ts` has a better pattern for this: read `previewDeposit` independently off `publicClient` and compare the CLI output to it. One assertion, oracle-free and independent of fork state, and it pins down the gateway, `tokenIn` and the decimals at once.

— Gabi 🤖

gndelia: Fixed in [06e0f603](https://github.com/vetro-protocol/vetro-monorepo/pull/821/commits/06e0f60370cbfc7464a316f9d30382fae39f2080). The test now reads `previewDeposit` on the gateway directly and expects the CLI to print the same value.

gabmontes (approved):

ArturDolzan (approved):

PR #823, opened 2026-09-15
Add daily TVL history to the subgraph

gndelia:
### Description

First PR of (hopefully) 3 parts to list historical TVL for our pegged tokens.

This PR updates the Vetro subgraph to track TVL on each gateway.

### Screenshots

Eventually (not this PR), we will build something in the analytics page that looks like this

image

The chart shows stacked bars listing to which stablecoin over the whole collateral they belong to

### Checklist

- [x] Manual testing passed.


subgraph/src/tvl.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  entity.save();
+
+  const tokens = whitelistedResult.value;
+  for (let i = 0; i < tokens.length; i++) {
+    const token = tokens[i];
+
+    const withdrawableResult = treasury.try_withdrawable(token);
+    if (withdrawableResult.reverted) {
+      continue;
```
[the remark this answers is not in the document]
gndelia: [Treasury.withdrawable](https://github.com/vetro-protocol/vetro-contracts/blob/3c063a03940de9aa39538e6a1bce0e8393ccb8e2/src/Treasury.sol#L424) does not revert itself. It only calls `convertToAssets` on an ERC4626 vault, which, if it follows the spec, doesn't revert either.

I don't think we'll add custom weird vaults - I think this is fine as is.

subgraph/schema.graphql
```diff
@@ -46,3 +46,21 @@ type VaultConfig @entity(immutable: true) {
...
+  totalSupply: BigInt! # Includes amoSupply
+}
+
+type TvlTokenHistory @entity(immutable: true) {
+  id: ID!
+  price: BigInt
+  tokenAddress: Bytes!
+  tvlHistory: TvlHistory!
+  unitPrice: BigInt
+  withdrawable: BigInt!
```
gabmontes: These fields carry units the schema doesn't state. `getPrice` is denominated in the gateway's peg unit, not USD: on the vetBTC treasury `getPrice(WBTC)` is `1e8` against a `unitPrice` of `1e8` (1.0 **BTC**), while `getPrice(USDT)` on the VUSD treasury is 0.99929 **USD**. So rows from the two gateways aren't in the same unit, and rendering vetBTC's history in USD needs a historical BTC/USD series this subgraph doesn't record.

— Auto Gabi

subgraph/src/tvl.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const tokenEntity = new TvlTokenHistory(`${id}-${token.toHexString()}`);
+    tokenEntity.tokenAddress = token;
+    tokenEntity.tvlHistory = id;
+    tokenEntity.withdrawable = withdrawableResult.value;
+
+    // Leave the price unset rather than skipping the token: the row is
+    // immutable and the day is never revisited, so its withdrawable would be
+    // lost for good.
+    const priceResult = treasury.try_getPrice(token);
```
gabmontes: Three valuation bases now disagree: `toTvlItems` (web) uses spot USD prices, `getTreasuryTotal` (api) uses the gateway's par redeem rate (`previewRedeem(USDT, 1e18)` is exactly `1000000`, `redeemFee` 0), and this stores the oracle mark (USDT 0.99929).

Storing the oracle price is right since spot can't be applied retroactively — but the chart's last bar won't match the collateral figure next to it, and the ~0.07% gap widens under a de-peg. Which basis will part 2 use?

— Gabi 🤖

subgraph/tests/tvl.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      "TvlTokenHistory",
+      usdcId,
+      "withdrawable",
+      usdcWithdrawable.toString(),
+    );
+    const stored = TvlTokenHistory.load(usdcId)!;
+    assert.assertTrue(!stored.isSet("price"));
+    assert.assertTrue(!stored.isSet("unitPrice"));
+
+    mockGetPrice(usdcAddress, BigInt.fromI32(99991000));
```
gabmontes: This re-mock (same at 318 and 337) only runs if every assertion above passes — `assert.fieldEquals` throws, and matchstick marks the test failed and continues with the rest of the file. `clearStore()` doesn't clear mocked functions (and 0.6.0 has no `clearMocks`), so a failure here leaves `getPrice(usdc)` reverting for every later test.

Current ordering absorbs it, but incidentally, and these lines read as dead code. Moving the default mocks into `beforeEach` would drop all three.

— Auto Gabi

subgraph/schema.graphql
gndelia: True, this is solved at the API level by calling the CMC REST APi and getting historical prices. Subgraphs don't allow calling external sources, so I can't solve it here. This is coming in a future PR

ArturDolzan (dismissed):

subgraph/src/tvl.ts
gndelia: Similar to https://github.com/vetro-protocol/vetro-monorepo/pull/823#discussion_r4021267652. - These values are not the final values exposed to the API, but they are crossed in the next PR  with CMC historical rates. There should not be problems against `toTvlItems` - all end up in USD values that come from CMC

subgraph/tests/tvl.test.ts
gndelia: Fixed in [651352b5](https://github.com/vetro-protocol/vetro-monorepo/pull/823/commits/651352b5ebffdd4a237e4c150632f4918411858f). Dropped all three restores rather than moving the defaults into `beforeEach`: every test already mocks each call it reads, so they stay order-independent, and `earn.test.ts` keeps the same shape (per-test mocks, `beforeEach` limited to `clearStore` + `setAddress`).

ArturDolzan (approved):

gabmontes (approved):

PR #824, opened 2026-09-15
Hide tooltip for zero amounts

### LINKED ISSUE #809: Hovering on a zero position shows the tooltip empty
### Description

Go to a Varible yield earn pool in which you don't have a position. It should show your position as "0"

Image

When hovering it, it shows an empty tooltip. If there's no value to show, it should show nothing on hover

Trynax:
### Description

Avoid wrapping zero token amounts in the tooltip component, preventing an empty tooltip from appearing for zero positions or balances.

### Screenshots

image

### Related issue(s)

Closes #809 

### Checklist

- [x] Manual testing passed.


ArturDolzan (dismissed):

gndelia (commented):
I think we can make a simpler approach. Isntead of changing `DisplayAmount`, we can pass to Tooltip a prop visible (or similar), based on being a non-zero amount

Then, `RcTooltip` accepts a `visible` prop that renders or hides the Tooltip based on this. I think this will make the whole PR far more simpler. Give it a try and let me know your thoughts

Trynax:
> I think we can make a simpler approach. Isntead of changing `DisplayAmount`, we can pass to Tooltip a prop visible (or similar), based on being a non-zero amount
> 
> Then, `RcTooltip` accepts a `visible` prop that renders or hides the Tooltip based on this. I think this will make the whole PR far more simpler. Give it a try and let me know your thoughts


Thanks, I tried this approach and pushed the update. `Tooltip` now accepts the `visible` prop, with zero amounts hidden.

gndelia (approved):

ArturDolzan (approved):

